Issue #46, opened 2024-07-30
`unmask_repos` doesn't ignore archived repositories


yegor256:
Currently, `unmask_repos` doesn't ignore archived repositories, but it should.

Yegorov:
@yegor256 may I take this one?

yegor256:
@Yegorov go ahead, please

PR #49, opened 2024-07-31
#46: ignore archived repositories for Fbe#unmask_repos

Yegorov:


lib/fbe/unmask_repos.rb
```diff
@@ -37,21 +37,23 @@ def self.mask_to_regex(mask)
...
-      Fbe.octo(loog:, global:, options:).repositories(mask.split('/')[0]).each do |r|
+      octo.repositories(mask.split('/')[0]).each do |r|
         repos << r[:full_name] if re.match?(r[:full_name])
       end
     end
     masks.select { |m| m.start_with?('-') }.each do |mask|
       re = Fbe.mask_to_regex(mask[1..])
       repos.reject! { |r| re.match?(r) }
     end
+    repos.reject! { |repo| octo.repository(repo)[:archived] }
```
yegor256, line 56: @Yegorov you are making an extra HTTP request to GitHub API. Maybe it's better to use the JSON already retrieved a few lines earlier, in line 48?

Yegorov: > Maybe it's better to use the JSON already retrieved a few lines earlier, in line 48?

@yegor256  I examined this case, but there may be a situation when an added repository without a mask is already in the archive

See this code: we added repos without make request to GitHub API:
```ruby
      unless mask.include?('*')
        repos << mask
        next
      end
``` 
Therefore, we should do an archive test at the very end

yegor256:
@Yegorov ok, thanks

0crat:
@Yegorov Thanks for the contribution! You've earned +15 points for this: +30 as a basis; -5 for very few (2) comments; -10 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running balance is [+76](https://zerocracy.github.io/judges-action/zerocracy-vitals.html).

0crat:
@yegor256 I've analyzed the work policy document and reviewed the draft message to the programmer. Here's a rewritten version of the message that's more conversational, incorporates relevant points from the policy, and aims to be more motivating:

Hey there! Thanks for jumping in on this review. I've got some good news and some room for improvement for you.

So, here's the scoop: You've earned 15 points for this review. Not bad, but let me break it down for you:

You started with a solid 25-point base - that's automatic for every review. Nice! 

Now, I noticed you only left one comment. The policy encourages more engagement, aiming for at least 6 comments to avoid a 10-point deduction. Unfortunately, we had to apply that deduction here.

But hey, look on the bright side! You didn't review your own code (that would've cost you 40 points), and you showed up to participate. That's great!

Want to boost your points next time? Here are some tips:
1. Leave more comments - you can earn up to 20 points for that!
2. If you're reviewing a larger piece of code (15+ hits-of-code), you'll avoid another 10-point deduction.
3. The more substantial the code review, the more points you can rack up (up to 40 points for the code size alone).

Your current balance is -579, but don't let that discourage you. Every review is a chance to climb up. Remember, you can earn up to 100 points per review!

Keep at it, and let's see those points soar. You've got this!

0crat:
@Yegorov Hey there! 👋 Thanks for the review, but reviewing your own code isn't the best practice. Your score for this one is +5 points: +25 base, -40 for self-review, -10 for only 3 comments (we like to see at least 6!), and +30 to keep you motivated. Remember, more comments and reviewing others' code can boost your score next time. Your total balance is now [+194](https://zerocracy.github.io/judges-action/zerocracy-vitals.html). Keep up the good work and let's aim for that 80-point max in future reviews! 💪

PR #203, opened 2025-06-07
#199: refactor ctor in `Fbe::Middleware::SqliteStore`

### LINKED ISSUE #199: The ctor of `Fbe::Middleware::SqliteStore` is not code-free
Currently, the ctor of `Fbe::Middleware::SqliteStore` is not code free, as it should be: https://www.yegor256.com/2015/05/07/ctors-must-be-code-free.html

Let's refactor the class to make sure we only work when methods are called, not when an object is created.

Yegorov:
Closes #199 

## Summary by CodeRabbit

- **Improvements**
	- Database connection is now established lazily on first use, improving resource management.
	- Cache table and index are created automatically during initial use.
	- Database close operation is deferred until program exit for safer cleanup.
- **Bug Fixes**
	- Enhanced error handling for invalid or corrupted database files and paths.
- **Tests**
	- Refactored tests for better reuse and clarity.
	- Added tests covering empty cache behavior, invalid database paths, corrupted database files, and deferred close callback verification.

### DIFF
--- a/lib/fbe/middleware/sqlite_store.rb
+++ b/lib/fbe/middleware/sqlite_store.rb
@@ -54,12 +54,6 @@ class Fbe::Middleware::SqliteStore
 
   private
 
-  def close
-    return unless @db
-    @db.close unless @db.closed?
-    @db = nil
-  end
-
   def perform(&)
     @db ||=
       SQLite3::Database.new(@path).tap do |d|
@@ -67,7 +61,7 @@ class Fbe::Middleware::SqliteStore
           t.execute 'CREATE TABLE IF NOT EXISTS cache(key TEXT UNIQUE NOT NULL, value TEXT);'
           t.execute 'CREATE INDEX IF NOT EXISTS key_idx ON cache(key);'
         end
-        at_exit { close }
+        at_exit { @db.close if @db && !@db.closed? }
       end
     @db.transaction(&)
   end
--- a/lib/fbe/middleware/sqlite_store.rb
+++ b/lib/fbe/middleware/sqlite_store.rb
@@ -17,13 +17,26 @@ class Fbe::Middleware::SqliteStore
   attr_reader :path
 
   def initialize(path)
-    raise ArgumentError, 'Database path cannot be nil or empty' if path.nil? || path.empty?
-    dir = File.dirname(path)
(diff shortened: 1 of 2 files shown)


yegor256:
@Yegorov looks over complicated to me, and very much imperative. Can't we do this instead:

```
def perform(&)
  @db ||= 
    begin
      # validate
      # open
      # prepare
    end
  @db.transaction(&)
end
```

Then, simply get rid of the `open` and `prepare` methods. WDYT?

yegor256:
@Yegorov also, we should have only the methods required by Faraday to be public in the class. If, for the example, the `close` method is not used by Faraday, we should keep it `private`.

Yegorov:
@yegor256 yes indeed, initially i added many public methods: `open`, `prepare`, `drop`, `close` and in the first version of refactoring I wanted to support them, which made the class itself more difficult.
Thanks for great articles! It's like a rabbit hole, there's a lot to think about.

Please check PR.

yegor256 (changes requested):
@Yegorov now it's much better! See a few ideas above.

lib/fbe/middleware/sqlite_store.rb
```diff
@@ -38,48 +35,34 @@ def delete(key)
...
   private
 
   def perform(&)
+    @db ||=
+      SQLite3::Database.new(@path).tap do |d|
+        d.transaction do |t|
+          t.execute 'CREATE TABLE IF NOT EXISTS cache(key TEXT UNIQUE NOT NULL, value TEXT);'
+          t.execute 'CREATE INDEX IF NOT EXISTS key_idx ON cache(key);'
+        end
+        at_exit { @db.close if @db && !@db.closed? }
```
yegor256, line 64: @Yegorov I can't imagine a situation when `@db` is nil or it's already closed. How can this happen?

Yegorov: @yegor256 the situation is extremely rare, but it can happen if after 64 line process receive signal `SIGTERM` (in this case `@db` not initialized yet). 
Сheck `!@db.closed?` really unnecessary.
Do you agree to leave only `if @db`?

Yegorov:
@yegor256 please check

yegor256:
@Yegorov thanks!

PR #242, opened 2025-07-06
#236: add test case that cache still available on duration of the age in octo client

### LINKED ISSUE #236: some URLs are not cached by SqliteStore
Look at this log: https://github.com/zerocracy/judges-action/actions/runs/16056495089/job/45312092276

```
/repositories/798641472: 1 (169ms)
```

I see this line twice, while it should only be seen once. This request should be cached. Let's find out what is the reason and fix.

Yegorov:
Closes #236 

## Summary by CodeRabbit

* **Tests**
  * Added a new test to verify that cached API responses remain available and valid for the specified duration before expiration.


Yegorov:
@yegor256 please check

PR #433, opened 2026-04-28
Fix #429: Raise descriptive error in Fbe::Graph#pull_requests_with_reviews when repository is missing

bibonix:
@yegor256, this PR fixes #429 and is ready for merge — all 10 CI checks pass.

## Problem

`Fbe::Graph#pull_requests_with_reviews` crashed with `NoMethodError: undefined method 'filter_map' for nil` whenever the GitHub GraphQL response did not include a `repository` key. This happens when the repository does not exist, the token lacks permission, or GraphQL returns a partial failure (`{"errors": [...]}`).

The crash originated at `lib/fbe/github_graph.rb:316`:

```ruby
.dig("repository", "pullRequests", "nodes")
.filter_map { ... }   # NoMethodError when result.dig returns nil
```

## Fix

Extract the `nodes` lookup before iterating, and raise a descriptive `Fbe::Error` when the repository is missing — the same pattern already used by `Fbe::Graph#total_commits` (`lib/fbe/github_graph.rb:141`). Callers now receive an actionable message such as `Repository 'owner/repo' not found` instead of an opaque `NoMethodError`.

## Tests

* Added `test_pull_requests_with_reviews_when_repository_is_missing` which stubs `#query` to return an errors-only response (matching the reproduction in the issue) and asserts an `Fbe::Error` whose message mentions the offending `owner/repo`. Without the production fix, this test fails with the exact `NoMethodError` reported in #429.
* All 248 pre-existing unit tests still pass; `bundle exec rake` (test, rubocop, picks, yard) is green locally.

## Commits

1. `#429: Add failing test reproducing NoMethodError when GraphQL response lacks repository key` — red test on master.
2. `#429: Raise descriptive Fbe::Error in pull_requests_with_reviews when repository key is missing` — production change that turns the test green.

Issue #429, opened 2026-04-26
NoMethodError in `pull_requests_with_reviews` when GitHub response has no `repository` key


morphqdd:
## What happens

`Fbe::Graph#pull_requests_with_reviews` crashes with `NoMethodError` when the GitHub GraphQL API returns a response without a `repository` key — which happens on permission errors, non-existent repositories, or partial GraphQL failures:

```
NoMethodError: undefined method 'filter_map' for nil
  lib/fbe/github_graph.rb:316:in 'Fbe::Graph#pull_requests_with_reviews'
```

The root cause is on line 315–316:

```ruby
'pulls_with_reviews' => result
  .dig('repository', 'pullRequests', 'nodes')
  .filter_map { ... }
```

When the API returns `{"errors": [{"message": "Could not resolve to a Repository"}]}`, `result.dig('repository', ...)` returns `nil`, and calling `.filter_map` on `nil` raises `NoMethodError`.

Minimal reproduction:

```ruby
graph = Fbe::Graph.new(token: 'fake')
def graph.query(_gql)
  { 'errors' => [{ 'message' => 'Could not resolve to a Repository' }] }
end
graph.pull_requests_with_reviews('bad-owner', 'bad-repo', Time.now - 86400)
# => NoMethodError: undefined method 'filter_map' for nil (github_graph.rb:316)
```

The same issue exists on line 323–324 for `has_next_page` and `next_cursor` via `.dig`.

## What should happen

When the API response does not contain a `repository` key, the method should raise a descriptive error or return an empty result — not crash with `NoMethodError`.

I would like to fix this issue if you confirm it's a bug.

morphqdd:
@yegor256, please take a look when you have a moment.

0crat:
@morphqdd Thanks for reporting a new bug! According to our team policy, you've earned **+18 points** for helping improve our project quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+72](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

bibonix:
[...]

PR #436, opened 2026-04-29
#434: Detect GraphQL errors centrally in Fbe::Graph#query

### LINKED ISSUE #434: Inconsistent error handling for missing repository in lib/fbe/github_graph.rb — direction?
@yegor256 — before sending another PR after #433, I'd like your direction on the broader pattern.

## What's there now

Two opposing philosophies coexist in `lib/fbe/github_graph.rb` for the same condition (GraphQL returns `data.repository = null` — token without access, repo deleted, partial errors, etc.):

**Crash** (your accepted pattern from #433):

| Method | Line | Pattern |
|---|---|---|
| `pull_requests_with_reviews` | 314 | `raise Fbe::Error if nodes.nil?` |

**Silent zero / empty** (4 places):

| Method | Line | Pattern |
|---|---|---|
| `total_issues_and_pulls` | 182-183 | `result.dig(...) \|\| 0` |
| `total_commits_pushed` | 437 | `commits.nil? ? 0 : ...` |
| `total_releases_published` | 507 | `break if releases.nil?` |
| `resolved_conversations` | 88-90 | `return [] if nodes.nil?` |

**Still crashing** (looks like #433 missed an analogous method 60 lines below):

| Method | Line | Issue |
|---|---|---|
| `pull_request_reviews` | 378 | `result['repository'].map` — same NoMethodError as #429 |
| `pull_request_reviews` | 382 | `v.dig('reviews', 'nodes').map` — partial GraphQL per-PR |

## Options for uniformity

**Option 1 — uniform raise (extend #433 to all methods)**
- Every method raises `Fbe::Error` on null repository
- Callers either rescue or rely on `Fbe.iterate` to log context and skip
- Pro: errors are visible, metrics aren't silently understated
- Con: callers must tolerate raise

**Option 2 — uniform silent (extend `|| 0` style to `pull_request_reviews`)**
- `pull_request_reviews` returns `[]` when repo missing
- All methods agree: "missing repo = no data, no error"
- Pro: simpler caller code, judge cycles survive inaccessible repos
- Con: metrics silently zero, no signal about "no access" vs "real zero"

**Option 3 — fix at the `query()` method (root cause)**
- `query()` (line 47-48) currently ignores `result.errors` — has the GraphQL error list but throws it away
- Make `query()` raise `Fbe::Error` with the GraphQL error message
- Downstream methods get clean data only, no per-method guards
- Pro: single point of correctness; each existing per-method nil-check becomes redundant and can be removed in follow-ups
- Con: bigger change, all callers must rescue or propagate

**Option 4 — extract a helper (`repo_or_raise(result, owner, name)` / `repo_or_default(...)`)**
- Centralise the choice in one helper; methods pick the variant
- Pro: explicit, testable, applied consistently
- Con: minor boilerplate

I'd suggest **Option 3** — it addresses the root cause and makes all four silent-zero methods correct by construction (they'll never receive null data). Ready to implement once you confirm direction.

vitaly-andr:
Per your "definitely the third option" reply on #434 — fixes the GraphQL error detection at the `query()` level instead of per-method.

`Fbe::Graph#query` was returning `result.data` without checking `result.errors`. When GitHub responds with a top-level `{"errors": [...]}` (token without access, repo deleted, etc.), `data` is `nil`, leading to silent zeros (`|| 0` in 4 methods) or downstream `NoMethodError` (the case #429 patched per-method).

Now `query()` raises `Fbe::Error` with the joined error messages whenever `result.errors` is non-empty. Per-method guards that protected against the same condition are removed:

- `pull_requests_with_reviews` — `raise if nodes.nil?` from #433
- `total_issues_and_pulls` — `|| 0` fallbacks (totalCount is `Int!`)
- `total_releases_published` — `releases.nil?` half of the break

Two guards are kept because they protect against legitimate non-error nulls (different concern from #434):

- `total_commits_pushed` — `defaultBranchRef` is null on a repo with no commits
- `resolved_conversations` — `pullRequest` is null when the PR doesn't exist

The existing test for the missing-repository case is updated to stub the GraphQL client (so the central check actually runs) instead of overriding `query()` itself.

Closes #434

### DIFF
--- a/lib/fbe/github_graph.rb
+++ b/lib/fbe/github_graph.rb
@@ -45,6 +45,10 @@ class Fbe::Graph # rubocop:disable Metrics/ClassLength
   #   puts result.viewer.login #=> "octocat"
   def query(qry)
     result = client.query(client.parse(qry))
+    if result.respond_to?(:errors) && !result.errors.empty?
+      msgs = result.errors.respond_to?(:messages) ? result.errors.messages.values.flatten : [result.errors.to_s]
+      raise(Fbe::Error, "GraphQL query failed: #{msgs.join('; ')}")
+    end
     result.data
   end
 
@@ -179,8 +183,8 @@ class Fbe::Graph # rubocop:disable Metrics/ClassLength
       GRAPHQL
     ).to_h
     {
-      'issues' => result.dig('repository', 'issues', 'totalCount') || 0,
-      'pulls' => result.dig('repository', 'pullRequests', 'totalCount') || 0
+      'issues' => result.dig('repository', 'issues', 'totalCount'),
+      'pulls' => result.dig('repository', 'pullRequests', 'totalCount')
     }
   end
 
@@ -311,7 +315,6 @@ class Fbe::Graph # rubocop:disable Metrics/ClassLength
       GRAPHQL
     ).to_h
     nodes = result.dig('repository', 'pullRequests', 'nodes')
-    raise(Fbe::Error, "Repository '#{owner}/#{name}' not found") if nodes.nil?
     {
       'pulls_with_reviews' => nodes.filter_map do |pull|
         next if pull.dig('timelineItems', 'nodes').empty?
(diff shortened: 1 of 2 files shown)


vitaly-andr:
@yegor256 ping per https://www.yegor256.com/2024/04/01/ping-me-please.html — open 4 days, CI green. Centralizes GraphQL error detection per #434. Could you review?

vitaly-andr:
Closing — withdrawing from contributing to zerocracy/* core repos due to a contribution-process ethics issue.

Over the past week, multiple PRs I opened on issues I reported went unreviewed while parallel PRs from other contributors were merged on the same issues — without those contributors leaving any review, comment, or coordination message on my open PRs, despite mine being visible in the issue's "Linked pull requests" sidebar for 47-66 hours before they opened their own. The architect merged those parallel PRs without acknowledging the originals. Three cases this week.

That's the system as designed. Not staying. Diff is public; if anyone wants the work, feel free.

PR #440, opened 2026-04-30
Mask Authorization header credentials in formatter logs (#439)

### LINKED ISSUE #439: `Authorization: Bearer ...` header logged in plaintext on 4xx/5xx responses
`Fbe::Middleware::Formatter` writes the full request `Authorization` header, including the live `Bearer <token>` value, into the configured logger whenever an upstream HTTP call returns 4xx or 5xx (anything that hits the 500/text or generic error branches at `lib/fbe/middleware/formatter.rb:70-96`).

The class docstring (lines 28-31) advertises the masked form `Authorization: "Bearer [FILTERED]"`, but no `Faraday::Logging::Formatter#filter` is ever registered — `lib/fbe/octo.rb:90` plugs the formatter in with `builder.use(Faraday::Response::Logger, loog, formatter: Fbe::Middleware::Formatter)` and stops there, so the inherited `apply_filters` is a no-op pass-through.

Reproduction (no fbe stubs, just the formatter on its own):

```ruby
require "faraday"
require "loog"
require "fbe/middleware/formatter"

loog = Loog::Buffer.new
formatter = Fbe::Middleware::Formatter.new(logger: loog, options: {})
formatter.request(
  Faraday::Env.from(
    method: :get,
    request_body: "some request body",
    url: URI("http://example.com"),
    request_headers: { "Authorization" => "Bearer github_pat_11AAsecret_REAL_TOKEN_VALUE" }
  )
)
formatter.response(
  Faraday::Env.from(status: 500, response_body: "Internal Server Error",
                    response_headers: { "content-type" => "text/html" })
)
puts loog.to_s
```

Output:

```
GET http://example.com HTTP/1.1
  Authorization: "Bearer github_pat_11AAsecret_REAL_TOKEN_VALUE"

  some request body
HTTP/1.1 500
  content-type: "text/html"

  Internal Server Error
```

## Why this matters even in GitHub Actions

Inside GitHub Actions, anything that comes through `${{ secrets.X }}` is auto-masked at log-write time, so the token shows up as `***` in the Actions UI. That hides the symptom but the root cause is in fbe and the leak is real for everyone else:

- Local development of fbe consumers — token visible in the terminal.
- Self-hosted runners or non-Actions CI (GitLab, Buildkite, Jenkins) without a global secret-masking layer.
- Any caller piping `loog` to a log shipper (Sentry, Loki, CloudWatch, Datadog), which records the plaintext before any UI-layer masking can kick in.
- judges-action's own `::add-mask::` (added in zerocracy/judges-action#1344) is an Actions-side mitigation; if for some reason the registered token differs from the one finally sent (env override, OAuth refresh, multi-token configurations), the masking misses.

## Existing test silently codifies the leak

`test/fbe/middleware/test_formatter.rb` uses a fake token `Bearer github_pat_11AAsecret` (line 94) and asserts `assert_match(/Authorization:/, str)` (lines 37, 67). The regex matches both the leaked and filtered forms, so the test passes whether the token is masked or not — it is not a meaningful security check.

## Proposed fix

Either:

1. Mask the value in `dump_headers`: when the header name is `Authorization` (case-insensitive), inspect only the scheme and replace the rest with `[FILTERED]`.
2. Register a `Faraday::Logging::Formatter#filter(/(Bearer)\s+\S+/i, '\1 [FILTERED]')` in `Fbe::Middleware::Formatter#initialize` so `apply_filters` actually does something.

Option 1 is local to `dump_headers` and covers any auth scheme (not just Bearer). Option 2 lines up with the docstring intent and keeps the filter list extensible.

Test side: replace `assert_match(/Authorization:/, str)` with the pair

```ruby
assert_match(/Authorization: "Bearer \[FILTERED\]"/, str)
refute_match(/github_pat_11AAsecret/, str)
```

so the suite catches future regressions.

vitaly-andr:
Fixes #439.

`Fbe::Middleware::Formatter` was logging the full `Authorization` header value, including the live `Bearer <token>`, on every 4xx/5xx response. The class docstring already promised `Authorization: "Bearer [FILTERED]"` but no `Faraday::Logging::Formatter#filter` was ever registered — `apply_filters` ran with an empty filter list and the token reached the logger verbatim.

## Fix

Override `Fbe::Middleware::Formatter#initialize` to register a regex filter that masks Bearer/Basic/Token schemes:

```ruby
def initialize(...)
  super
  filter(/(Authorization:\s*"?(?:Bearer|Basic|Token))\s+[^"\s]+/i, '\1 [FILTERED]')
end
```

The regex anchors on the literal `Authorization:` prefix (with optional whitespace and opening quote from `dump_headers`'s `inspect`), captures the auth scheme, then replaces the credential portion with `[FILTERED]`. This avoids over-matching free-form `Bearer` mentions in URL paths or response bodies.

## Test changes

The existing assertions used `assert_match(/Authorization:/, str)`, which matched both the leaked and the masked form — passing whether the token was redacted or not. Replaced with the stronger pair:

```ruby
assert_match(/Authorization: "Bearer \[FILTERED\]"/, str)
refute_match(/github_pat_11AAsecret/, str)
```

so future regressions in the masking path will be caught.

`bundle exec rake` clean locally: 5 formatter tests / 40 assertions, 0 failures; 65 files rubocop clean.

### DIFF
--- a/lib/fbe/middleware/formatter.rb
+++ b/lib/fbe/middleware/formatter.rb
@@ -39,6 +39,15 @@ require_relative '../../fbe/middleware'
 # Copyright:: Copyright (c) 2024-2026 Zerocracy
 # License:: MIT
 class Fbe::Middleware::Formatter < Faraday::Logging::Formatter
+  # Registers regex filters that mask credentials in Authorization headers
+  # before they reach the underlying logger. The Faraday::Logging::Formatter
+  # base class invokes #apply_filters on every logged string; without this
+  # override the inherited filter list is empty and tokens leak verbatim.
+  def initialize(...)
+    super
+    filter(/(Authorization:\s*"?(?:Bearer|Basic|Token))\s+[^"\s]+/i, '\1 [FILTERED]')
+  end
+
   # Captures HTTP request details for later use in error logging.
   #
   # @param [Hash] http Request data including method, url, headers, and body
--- a/test/fbe/middleware/test_formatter.rb
+++ b/test/fbe/middleware/test_formatter.rb
@@ -34,12 +34,13 @@ class LoggingFormatterTest < Fbe::Test
       refute_empty(str)
       [
         %r{http://example.com},
-        /Authorization:/,
+        /Authorization: "Bearer \[FILTERED\]"/,
         %r{HTTP/1.1 401},
         /x-github-api-version-selected: "2022-11-28"/,
         /hello, world!/,
         /request body/
       ].each { |ptn| assert_match(ptn, str) }
+      refute_match(/github_pat_11AAsecret/, str)
     end
   end
 
@@ -64,13 +65,14 @@ class LoggingFormatterTest < Fbe::Test
       refute_empty(str)
       [
         %r{http://example.com},
-        /Authorization:/,
+        /Authorization: "Bearer \[FILTERED\]"/,
         /some request body/,
         %r{HTTP/1.1 502},
         /x-github-api-version-selected: "2022-11-28"/,
         %r{content-type: "text/html; charset=utf-8"},
         "#{body.slice(0, 97)}..."
       ].each { |ptn| assert_match(ptn, str) }
+      refute_match(/github_pat_11AAsecret/, str)
     end
   end
 


vitaly-andr:
@yegor256 ping per https://www.yegor256.com/2024/04/01/ping-me-please.html — open 3 days, CI green, MERGEABLE/CLEAN. Could you review when you get a chance?

vitaly-andr:
Closing — withdrawing from contributing to zerocracy/* core repos due to a contribution-process ethics issue.

Over the past week, multiple PRs I opened on issues I reported went unreviewed while parallel PRs from other contributors were merged on the same issues — without those contributors leaving any review, comment, or coordination message on my open PRs, despite mine being visible in the issue's "Linked pull requests" sidebar for 47-66 hours before they opened their own. The architect merged those parallel PRs without acknowledging the originals. Three cases this week.

That's the system as designed. Not staying. Diff is public; if anyone wants the work, feel free.

PR #444, opened 2026-05-01
#442: Omit GraphQL after argument when cursor is nil

### LINKED ISSUE #442: `pull_requests_with_reviews` sends `after: ""` instead of omitting `after` when cursor is `nil`
## What Happens

`pull_requests_with_reviews` and `pull_request_reviews` in `lib/fbe/github_graph.rb`
interpolate the cursor directly into the GraphQL query:

```ruby
pullRequests(first: 100, after: "#{cursor}")
```

When `cursor` is `nil` (first page), Ruby string interpolation produces `after: ""`
(empty string), not a valid GraphQL cursor. GitHub API may reject this with an error
or return unexpected results — an empty string is not a valid opaque cursor value.

## What Should Happen

When `cursor` is `nil`, the `after` argument should be omitted entirely, as done
correctly in `total_commits_pushed` (line 405):

```ruby
after = "after: \"#{cursor}\", " unless cursor.nil?
# then use #{after} in the query
```

Both `pull_requests_with_reviews` (line 294) and `pull_request_reviews` (line 356)
need the same fix.

bibonix:
@yegor256 this PR fixes #442.

**The bug.** Both `pull_requests_with_reviews` (line 294) and `pull_request_reviews` (line 356) in `lib/fbe/github_graph.rb` interpolated the `cursor` value directly into the GraphQL query as `after: "#{cursor}"`. When the caller passed `cursor: nil` (for the first page), Ruby produced `after: ""` — an empty string, which is not a valid opaque cursor and is rejected/misinterpreted by the GitHub GraphQL API.

**The fix.** I mirrored the pattern already used correctly in `total_commits_pushed` (line 405): build an `after = "after: "#{cursor}", " unless cursor.nil?` snippet and interpolate it into the query. When `cursor` is `nil`, `after` is `nil` and the `after:` argument is omitted from the query entirely. When `cursor` is a real cursor, the argument is rendered as before. This required a one-line change inside each of the two methods.

**Tests.** I added two tests in `test/fbe/test_github_graph.rb` that capture the GraphQL query string sent through `Fbe::Graph#query` and assert it does not contain `after: ""` when `cursor` is `nil`. Both tests fail on `master` and pass with this change. The full test suite (`bundle exec rake test`) and `rubocop` both pass locally and in CI.

Ready for merge.

### DIFF
--- a/lib/fbe/github_graph.rb
+++ b/lib/fbe/github_graph.rb
@@ -287,11 +287,12 @@ class Fbe::Graph # rubocop:disable Metrics/ClassLength
   #     cursor = json['next_cursor']
   #   end
   def pull_requests_with_reviews(owner, name, since, cursor: nil)
+    after = "after: \"#{cursor}\", " unless cursor.nil?
     result = query(
       <<~GRAPHQL
         {
           repository(owner: "#{owner}", name: "#{name}") {
-            pullRequests(first: 100, after: "#{cursor}") {
+            pullRequests(#{after}first: 100) {
               nodes {
                 id
                 number
@@ -349,11 +350,12 @@ class Fbe::Graph # rubocop:disable Metrics/ClassLength
   def pull_request_reviews(owner, name, pulls: [])
     requests =
       pulls.map do |number, cursor|
+        after = "after: \"#{cursor}\", " unless cursor.nil?
         <<~GRAPHQL
           pr_#{number}: pullRequest(number: #{number}) {
             id
             number
-            reviews(first: 100, after: "#{cursor}") {
+            reviews(#{after}first: 100) {
               nodes {
                 id
                 submittedAt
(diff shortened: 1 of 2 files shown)


0crat:
@bibonix Hey! Nice work on that contribution – you just scored the full **+24 points**! 🎉 Your code hit the sweet spot with good size and solid review feedback, which is exactly how our [bonus system](https://www.yegor256.com/2014/11/14/hits-of-code.html) rewards quality work. [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming and you'll see those points add up fast! Your running score is [+180](https://zerocracy.github.io/judges-action/zerocracy-vitals.html), so don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for the latest updates.

PR #448, opened 2026-05-03
#447: make off_quota? aware of GitHub Search API quota

### LINKED ISSUE #447: `Fbe.octo.off_quota?` ignores GitHub Search API quota — `search_issues` hits 403 even with core quota available
The `off_quota?` method in `lib/fbe/octo.rb:157` only checks `rate_limit.remaining`, which reflects GitHub's **core quota** (5,000 req/hr). GitHub maintains a **separate Search API quota** of 30 requests per minute (1,000/hr), which `off_quota?` never inspects.

This causes downstream judges (e.g., `quality-of-service` in `judges-action`, which does 7 distinct `Fbe.octo.search_issues` calls per slice) to repeatedly issue search requests after the search quota is exhausted, getting `403 / API rate limit exceeded` even though the core quota guard reports thousands of requests remaining.

Real-world hit: zerocracy/judges-action#589 — `quality-of-service` looped on a search query and got four consecutive 403s within 30 seconds:

```
GET https://api.github.com/search/issues?...&q=repo:objectionary/eolang-paper+type:issue+created:2024-12-26..2025-01-25 -> 403 / API rate limit exceeded
GET (same URL) -> 403 ...
GET (same URL) -> 403 ...
GET (same URL) -> 403 ...
```

**Suggested fix**

Extend the off-quota guard so that for `search_*` methods (`search_issues`, `search_commits`, `search_repositories`, `search_users`, `search_code`) it consults the Search resource of `/rate_limit` (`body[:resources][:search]`) instead of the core resource. Calls to non-search methods continue to use the existing core check.

The Faraday rate-limit middleware (`lib/fbe/middleware/rate_limit.rb`) caches only `rate.remaining` (core); it should additionally cache `resources.search.remaining` so the guard can be cheap.

Reference: https://docs.github.com/en/rest/rate-limit/rate-limit

igor-belousov:
Closes #447. Downstream consumer: zerocracy/judges-action#589.

GitHub has two independent rate limits: core (5000/hr) and search (30/min).
`off_quota?` previously only checked core, so search endpoints proceeded
even when search quota was exhausted, then crashed mid-judge with 403.

Changes:
* `Fbe::SEARCH_METHODS` lists the 6 Octokit search methods.
* `off_quota?` accepts `resource: :core | :search` and `threshold:`
  (defaults: 50 for core, 5 for search). For `:search` it reads
  `resources.search.remaining` from `last_response`, falls back to core
  with a warning if the search payload is unavailable.
* The `intercepted` block routes `SEARCH_METHODS` calls through the
  search-quota check.
* `Fbe::Middleware::RateLimit` now tracks `@searchleft` separately and
  decrements it for `/search/*` paths, so the cached `/rate_limit`
  response reflects accurate search-remaining counts within the
  100-call cache window.
* The cache-hit path no longer mutates the cached response body. It
  builds a fresh body via `patched_body` (deep-dup via JSON round-trip)
  and a `response_env` that dups the original Faraday env, preserving
  invariants like `finished?`.

Tests: 11 new tests across `test/fbe/test_octo.rb` and
`test/fbe/middleware/test_rate_limit.rb` covering quota routing,
threshold defaults, fallback paths, cached-stale scenario across the
100-request window, the cache-body-leak regression, and the
JSON-string body decrement path. 268 tests total, all green, rubocop
clean.

Known follow-ups (separate issues to be filed):
* `.send` / `.public_send` on the `Fbe.octo` wrapper bypasses the
  quota guard. The wrapper extends `BasicObject`, so `:send` lands
  in `method_missing` with `m = :send`. Affects both core and
  search paths, predates this change.
* Source-of-truth: middleware tracks `@searchleft` while `off_quota?`
  reads `last_response`. Should consolidate so middleware is the
  single source. Current design works but is fragile (depends on
  call ordering).
* Concurrency hardening: middleware state is not thread-safe.
  `judges-action` runs sequentially today, so not an immediate issue,
  but should be guarded with a mutex if concurrent callers are ever
  introduced. Marked with an inline NOT thread-safe comment.

### DIFF
--- a/lib/fbe/middleware/rate_limit.rb
+++ b/lib/fbe/middleware/rate_limit.rb
@@ -84,7 +84,8 @@ class Fbe::Middleware::RateLimit < Faraday::Middleware
   # @param [Faraday::Response] response The API response
   # @return [Integer] The remaining requests count
   def extract_remaining_count(response)
-    body = parsed_body(response)
+    body = response.body
+    body = JSON.parse(body) if body.is_a?(String)
     return 0 unless body.is_a?(Hash)
     body.dig('rate', 'remaining') || 0
   end
@@ -94,21 +95,10 @@ class Fbe::Middleware::RateLimit < Faraday::Middleware
   # @param [Faraday::Response] response The API response
   # @return [Integer, nil] Remaining search-API requests or nil if absent
   def extract_search_remaining_count(response)
-    body = parsed_body(response)
+    body = response.body
+    body = JSON.parse(body) if body.is_a?(String)
     return nil unless body.is_a?(Hash)
     body.dig('resources', 'search', 'remaining')
-  end
-
-  # Parses the response body as a Hash if it's a JSON string.
-  #
-  # @param [Faraday::Response] response The API response
-  # @return [Hash, Object] Parsed body or original on parse failure
-  def parsed_body(response)
-    body = response.body
-    return body unless body.is_a?(String)
-    JSON.parse(body)
-  rescue JSON::ParserError
-    nil
   end
 
   # Builds a fresh body with the current remaining counts written in,
@@ -119,17 +109,14 @@ class Fbe::Middleware::RateLimit < Faraday::Middleware
   # @return [Object] A new body of the same type with remaining counts updated
   def patched_body(original)
     stringed = original.is_a?(String)
-    if stringed
-      begin
-        body = JSON.parse(original)
-      rescue JSON::ParserError
+    body =
+      if stringed
+        JSON.parse(original)
+      elsif original.is_a?(Hash)
+        JSON.parse(original.to_json)
+      else
         return original
       end
-    elsif original.is_a?(Hash)
-      body = JSON.parse(original.to_json)
-    else
-      return original
-    end
     body['rate']['remaining'] = @remaining if body['rate']
     body.dig('resources', 'search')&.[]=('remaining', @searchleft) if @searchleft
     stringed ? body.to_json : body
--- a/lib/fbe/middleware/rate_limit.rb
+++ b/lib/fbe/middleware/rate_limit.rb
@@ -24,6 +24,8 @@ require_relative '../../fbe/middleware'
 # Copyright:: Copyright (c) 2024-2026 Zerocracy
 # License:: MIT
 class Fbe::Middleware::RateLimit < Faraday::Middleware
+  # NOT thread-safe: assumes single-threaded use (judges run sequentially in judges-action).
+  #
   # Initializes the rate limit middleware.
   #
   # @param [Object] app The next middleware in the stack
@@ -31,6 +33,7 @@ class Fbe::Middleware::RateLimit < Faraday::Middleware
     super
     @cached = nil
     @remaining = nil
+    @searchleft = nil
     @counter = 0
   end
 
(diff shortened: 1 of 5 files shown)


igor-belousov:
@yegor256 review please

lib/fbe/middleware/rate_limit.rb
```diff
@@ -56,76 +59,95 @@ def call(env)
...
-  # @param [Faraday::Response] response The cached response to update
-  def update_remaining_count(response)
+  # @param [Faraday::Response] response The API response
+  # @return [Hash, Object] Parsed body or original on parse failure
+  def parsed_body(response)
     body = response.body
-    stringed = body.is_a?(String)
+    return body unless body.is_a?(String)
+    JSON.parse(body)
+  rescue JSON::ParserError
```
yegor256: @igor-belousov two terrible design ideas in one line: exception swallowing and returning NULL

igor-belousov: @yegor256 the rescue is there because test_ignores_non_hash_response_body on master feeds 'invalid json' to /rate_limit and expects the middleware to pass it through. Without rescue, that test crashes. Want me to delete the test, or keep one rescue at the call site that logs and raises Fbe::Error?

yegor256: @igor-belousov 

> feeds 'invalid json' to /rate_limit and expects the middleware to pass it through

why should it pass an invalid JSON through?

igor-belousov: Fair point, removing the test and inlining JSON.parse without rescue

igor-belousov:
@yegor256 pushed cleanup in 82fe71f, dropped the rescue and the test. Ready for another look.

lib/fbe/middleware/rate_limit.rb
```diff
@@ -80,52 +85,56 @@ def track_request
...
+  def extract_search_remaining_count(response)
     body = response.body
-    stringed = body.is_a?(String)
-    if stringed
-      begin
-        body = JSON.parse(body)
-      rescue JSON::ParserError
-        return
+    body = JSON.parse(body) if body.is_a?(String)
+    return nil unless body.is_a?(Hash)
```
yegor256: @igor-belousov again, why returning `nil` instead of zero. In general, `nil` is a bad practice in programming, stay away from it.

igor-belousov:
@yegor256 fixed in 8b52868, replaced nil with 0 for both counters. Ready for another look.

yegor256:
@rultor merge

0crat:
@yegor256 Thanks for the review! You've earned **+19 points** according to our policy: **+18** base points for reviewing, **+6** for the 420 hits-of-code (0.015 × 420, capped at 16), and **-5** deducted for having fewer than 9 comments during review (only 1 comment made). Your running score is [+633](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too! 🎯

0crat:
@igor-belousov Hey there! Nice work on your contribution – you've bagged **+20 points** this time! 🎉 Here's the breakdown: started with **+24** as the base reward, then had to dock **-4** since your **420** hits-of-code went over our **133** threshold (remember, we're aiming for quality over quantity!). Your running score is looking solid at **[+212](https://zerocracy.github.io/judges-action/zerocracy-vitals.html)** – [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) those contributions flowing, and don't forget to peek at your **[Zerocracy account](https://www.zerocracy.com/)** for the latest updates!

PR #462, opened 2026-05-15
#457: use judge parameter instead of $judge global in log messages

### LINKED ISSUE #457: regularly.rb logs use $judge global instead of the judge parameter
In `lib/fbe/regularly.rb` the two log messages reference `$judge` (the global) while the fact is correctly stored using the `judge` keyword parameter:

```ruby
loog.info("#{$judge} statistics were collected #{recent.when.ago} ago, skipping...")
loog.info("#{$judge} statistics weren't collected for the last #{interval} days")
# but then...
f.what = judge   # uses the parameter, not the global
```

If the method is called with a custom `judge:` argument (e.g. in tests or multi-judge setups), the log lines will display the wrong judge name. The stored fact is correct; only the logging is wrong.

Fix: replace `$judge` with `judge` in both `loog.info` calls on those two lines.

morphqdd:
Fixes #457.

Replaced `$judge` global with the `judge` keyword parameter in two `loog.info` calls inside `Fbe.regularly`. The fact insertion already used the parameter correctly (`f.what = judge`); only the log messages were inconsistent. With this fix, callers that pass an explicit `judge:` argument (tests, multi-judge setups) get correct judge names in the log output.

Added regression test `test_log_uses_judge_parameter_not_global` that sets `$judge` to one value, calls `Fbe.regularly` with a different `judge:`, captures output via `Loog::Buffer`, and asserts the parameter value (not the global) appears in the logs.

### DIFF
--- a/lib/fbe/regularly.rb
+++ b/lib/fbe/regularly.rb
@@ -44,12 +44,12 @@ def Fbe.regularly(area, p_every_days, p_since_days = nil, fb: Fbe.fb, judge: $ju
   ).each.first
   if recent
     loog.info(
-      "#{$judge} statistics were collected #{recent.when.ago} ago, " \
+      "#{judge} statistics were collected #{recent.when.ago} ago, " \
       "skipping now (we run it every #{interval} days)"
     )
     return
   end
-  loog.info("#{$judge} statistics weren't collected for the last #{interval} days")
+  loog.info("#{judge} statistics weren't collected for the last #{interval} days")
   fb.txn do |fbt|
     f = fbt.insert
     f.what = judge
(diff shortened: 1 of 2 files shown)


morphqdd:
@yegor256, please take a look when you have a moment.

0crat:
@morphqdd Thanks for the contribution! You've earned **+20 points** for this: +24 as a basis; -4 for contributing fewer than 30 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (20 total). Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+689](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #470, opened 2026-05-20
#451 guard Fbe::Middleware::RateLimit state with a mutex

### LINKED ISSUE #451: Fbe::Middleware::RateLimit is not thread-safe
## What

`Fbe::Middleware::RateLimit` keeps mutable state (`@cached`, `@remaining`, `@searchleft`, `@counter`) in plain instance variables and updates them with non-atomic compound operations like `@counter += 1` and `@remaining -= 1`. The middleware is shared across all requests through the same Faraday connection, so any concurrent caller would race on these vars.

## Why it's not breaking today

`zerocracy/judges-action` runs judges sequentially in a single process. There is no concurrent caller against the same `Fbe.octo` instance, so the middleware works correctly in practice.

The middleware file carries a comment marking the assumption explicitly:

> NOT thread-safe: assumes single-threaded use (judges run sequentially in judges-action).

## Why this matters for the future

Plausible future scenarios that would break:

1. Parallel judges (e.g. running several scorers concurrently to speed up a long pipeline).
2. Async / fiber-based fetches inside a single judge.
3. Any caller outside `judges-action` that uses `Fbe.octo` from threads (e.g. a Sinatra app, a worker pool).

In all three cases the symptoms would be hard to debug:
- Lost decrements on `@counter` would delay the 100-request refresh, potentially serving stale cached `/rate_limit` for much longer than intended.
- Lost decrements on `@searchleft` / `@remaining` would over-report remaining quota, causing `off_quota?` to let calls through that should have been blocked, leading to mid-request 403s (the exact failure mode #447 was trying to prevent).
- `handle_rate_limit_request` could race two threads through the cache-miss branch, both calling `@app.call(env)` and clobbering `@cached`.

## Proposed fix

Wrap mutating sections in a `Mutex#synchronize` block. Cost is negligible compared to the cost of an HTTP call. Sketch:

```ruby
def initialize(app)
  super
  @lock = Mutex.new
  # ...
end

def call(env)
  if env.url.path == '/rate_limit'
    @lock.synchronize { handle_rate_limit_request(env) }
  else
    @lock.synchronize { track_request(env.url.path) }
    @app.call(env)
  end
end
```

(Note: `@app.call(env)` should NOT be inside the lock or it serializes all requests.)

## Why this wasn't done in #447

Out of scope for the search-quota fix. Marked with the inline `NOT thread-safe` comment so the assumption is explicit until this issue is addressed.

Filed during PR #448.

edmoffo:
Closes #451.

`Fbe::Middleware::RateLimit` keeps mutable state (`@cached`, `@remaining`, `@searchleft`, `@counter`) in plain instance variables and updates them with non-atomic compound operations like `@counter += 1` and `@remaining -= 1`. The middleware instance is shared across every request through the same Faraday connection, so any concurrent caller (parallel judges, fiber-based fetches, anything that uses `Fbe.octo` from threads) would race on these vars and lose decrements, serve stale cached `/rate_limit` for longer than the 100-request window intended, and let two threads through the cache-miss branch at the same time.

The change adds a `Mutex` around the cache-refresh path in `handle_rate_limit_request` and around the counter mutation in `track_request`, while keeping `@app.call(env)` outside the lock so a slow downstream request does not serialize unrelated callers. The "NOT thread-safe" header comment that documented the prior limitation is gone.

- `bundle exec rake test TEST=test/fbe/middleware/test_rate_limit.rb` — 16 tests, 32 assertions, 0 failures, 0 errors, including the two new regression tests `test_concurrent_non_rate_limit_requests_do_not_lose_decrements` and `test_concurrent_rate_limit_requests_refresh_once`.
- `bundle exec rake test` — 275 tests, 760 assertions, 0 new failures (10 pre-existing errors in `test_octo.rb` and adjacent quota-dependent tests reproduce on `master` in this environment and are unrelated to this change).
- `bundle exec rake rubocop` — 65 files inspected, no offenses.
- `bundle exec rake yard` — 81.82% documented, no new warnings.

### DIFF
--- a/lib/fbe/middleware/rate_limit.rb
+++ b/lib/fbe/middleware/rate_limit.rb
@@ -24,8 +24,6 @@ require_relative '../../fbe/middleware'
 # Copyright:: Copyright (c) 2024-2026 Zerocracy
 # License:: MIT
 class Fbe::Middleware::RateLimit < Faraday::Middleware
-  # NOT thread-safe: assumes single-threaded use (judges run sequentially in judges-action).
-  #
   # Initializes the rate limit middleware.
   #
   # @param [Object] app The next middleware in the stack
@@ -35,6 +33,7 @@ class Fbe::Middleware::RateLimit < Faraday::Middleware
     @remaining = 0
     @searchleft = 0
     @counter = 0
+    @lock = Mutex.new
   end
 
   # Processes the HTTP request and handles rate limit caching.
@@ -43,9 +42,9 @@ class Fbe::Middleware::RateLimit < Faraday::Middleware
   # @return [Faraday::Response] The response from cache or the next middleware
   def call(env)
     if env.url.path == '/rate_limit'
-      handle_rate_limit_request(env)
+      @lock.synchronize { handle_rate_limit_request(env) }
     else
-      track_request(env.url.path)
+      @lock.synchronize { track_request(env.url.path) }
       @app.call(env)
     end
   end
--- a/test/fbe/middleware/test_rate_limit.rb
+++ b/test/fbe/middleware/test_rate_limit.rb
@@ -236,6 +236,47 @@ class RateLimitTest < Fbe::Test
     assert_equal(4999, parsed['rate']['remaining'])
   end
 
+  def test_concurrent_non_rate_limit_requests_do_not_lose_decrements
+    payload = {
+      'rate' => { 'limit' => 5000, 'remaining' => 10_000, 'reset' => 1_672_531_200 },
+      'resources' => {
+        'core' => { 'limit' => 5000, 'remaining' => 10_000, 'reset' => 1_672_531_200 },
+        'search' => { 'limit' => 30, 'remaining' => 1_000, 'reset' => 1_672_531_200 }
+      }
+    }
+    stub_request(:get, 'https://api.github.com/rate_limit')
+      .to_return(status: 200, body: payload.to_json, headers: { 'Content-Type' => 'application/json' })
+    stub_request(:get, 'https://api.github.com/user')
+      .to_return(status: 200, body: '{"login":"x"}', headers: { 'Content-Type' => 'application/json' })
+    stub_request(:get, 'https://api.github.com/search/issues?q=z')
+      .to_return(status: 200, body: '{"items":[]}', headers: { 'Content-Type' => 'application/json' })
+    conn = create_connection
+    conn.get('/rate_limit')
+    threads =
+      Array.new(10) do
+        Thread.new do
+          4.times do
+            conn.get('/user')
+            conn.get('/search/issues?q=z')
+          end
+        end
+      end
+    threads.each(&:join)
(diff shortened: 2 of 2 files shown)


bibonix (approved):
The mutex wrap is the minimal correct fix for #451:  and  both mutate the same four instance variables (`@cached`, `@remaining`, `@searchleft`, `@counter`), and pulling both call sites under one `@lock.synchronize` block makes the compound updates atomic without leaking the lock over `@app.call(env)` on the hot non-rate_limit path. The two new tests pin the two regressions that the lock prevents — counter loss across concurrent non-rate_limit calls, and duplicated refresh calls to `/rate_limit` — and the existing 14 tests in the file still pass. The "NOT thread-safe" docstring is correctly removed now that the contract has changed. CI is green on all jobs.

Worth noting for the next reader: the rate_limit branch does hold the lock across `@app.call(env)` so concurrent refresh callers coalesce into a single upstream HTTP call (which is the point of `test_concurrent_rate_limit_requests_refresh_once`), so the PR body's claim that `@app.call(env)` stays outside the lock only applies to the non-rate_limit branch. The behavior is correct; the description just covers half of it.

bibonix (approved):
The mutex wrap is the minimal correct fix for #451: `handle_rate_limit_request` and `track_request` both mutate the same four instance variables (`@cached`, `@remaining`, `@searchleft`, `@counter`), and pulling both call sites under one `@lock.synchronize` block makes the compound updates atomic without leaking the lock over `@app.call(env)` on the hot non-rate_limit path. The two new tests pin the two regressions that the lock prevents — counter loss across concurrent non-rate_limit calls, and duplicated refresh calls to `/rate_limit` — and the existing 14 tests in the file still pass. The "NOT thread-safe" docstring is correctly removed now that the contract has changed. CI is green on all jobs.

Worth noting for the next reader: the rate_limit branch does hold the lock across `@app.call(env)` so concurrent refresh callers coalesce into a single upstream HTTP call (which is the point of `test_concurrent_rate_limit_requests_refresh_once`), so the PR body's claim that `@app.call(env)` stays outside the lock only applies to the non-rate_limit branch. The behavior is correct; the description just covers half of it.

0crat:
@bibonix **Great work on the review!** 🎉 You've earned **+10 points** (+18 base, -8 for missing comments). Your total score is now [+662](https://zerocracy.github.io/judges-action/zerocracy-vitals.html). Remember to check your [Zerocracy account](https://www.zerocracy.com/) for updates!

0crat:
@edmoffo Hey! Nice work on that contribution – you just scored the full **+24 points**! 🎯 Your code hit the sweet spot with good size, solid reviews, and clean feedback. [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) this momentum going and you'll keep maxing out those points. Your running score is [+689](https://zerocracy.github.io/judges-action/zerocracy-vitals.html) – don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too!

PR #475, opened 2026-05-24
#449 route Fbe.octo's send and public_send through the quota guard

### LINKED ISSUE #449: Fbe.octo wrapper: .send / .public_send bypasses the off_quota? guard
## What

Calling `.send(:method_name, args)` or `.public_send(:method_name, args)` on the object returned by `Fbe.octo` bypasses the pre-call quota guard entirely. The call reaches the underlying Octokit client directly without going through the `off_quota?` check.

## Why it happens

`Fbe.octo` wraps the Octokit client with `intercepted` (from the `intercepted` gem), which builds a `BasicObject` subclass and routes everything through `method_missing`. `BasicObject` does not define `send` or `public_send`, so:

1. `wrapper.send(:search_issues, 'q')` triggers `method_missing(:send, :search_issues, 'q')`.
2. The intercepted block sees `m = :send`, which is NOT in `Fbe::SEARCH_METHODS` and is not `off_quota?`.
3. Falls through to `@o.__send__(:send, :search_issues, 'q')` on the inner decoor object, which dispatches `search_issues` directly without the quota check.

Net effect: any caller using `.send` / `.public_send` on `Fbe.octo` (intentionally or via a meta-programming helper) skips the entire quota protection introduced by #447 and the original core check.

## How I noticed

Discovered while writing tests for #447. The test `test_search_methods_are_routed_through_search_quota_check` initially used `o.public_send(m, 'q')` to iterate over `Fbe::SEARCH_METHODS` and assert each was blocked. The test failed for every method because `public_send` itself was being intercepted instead. Switched to `o.__send__(m, 'q')` to make the test work.

## Reproduction

```ruby
o = Fbe.octo(loog: Loog::NULL, global: {}, options: Judges::Options.new)
# search quota is exhausted (remaining: 1, threshold: 5)
o.search_issues('q')           # raises Fbe::OffQuota (correct)
o.send(:search_issues, 'q')    # makes the HTTP call (bug)
o.public_send(:search_issues, 'q')  # also makes the HTTP call (bug)
o.__send__(:search_issues, 'q')     # raises Fbe::OffQuota (correct)
```

## Scope

Affects both core and search quota paths. Predates #447 / PR #448; the search work just made it visible.

## Possible fixes

1. Define `send` and `public_send` on the intercepted wrapper that re-dispatch through the wrapper's own `method_missing`. Simplest, but requires upstream change to the `intercepted` gem (or a local override).
2. Document that `Fbe.octo` callers must use direct method invocation, never `.send` / `.public_send`. Cheapest, but easy to forget.
3. Replace `intercepted` with a different wrapping strategy that doesn't inherit from `BasicObject`.

Discovered while working on #447 / PR #448.

bibonix:
Closes #449.

`Fbe.octo` wraps Octokit with the `intercepted` gem, which builds a `BasicObject` subclass and routes every call through `method_missing`. Neither `send` nor `public_send` is defined on `BasicObject`, so `o.send(:search_issues, 'q')` arrived at the interceptor block with `m=:send`, fell through the `Fbe::SEARCH_METHODS` and `off_quota?` guard list, and dispatched the underlying `search_issues` directly — the search-quota check from #447 never ran. `public_send` had the same problem. The repro from #449 (`o.send(:search_issues, 'q')` while search is exhausted) reached the GitHub API instead of raising `Fbe::OffQuota`.

Define `send` and `public_send` on the intercepted wrapper that forward to `__send__`. Both forwarders use Ruby's `...` shorthand to preserve positional, keyword, and block arguments. `__send__` on the wrapper hits the dispatcher that already triggers the interceptor block, so the real method name now reaches the guard and the existing quota logic raises as designed for both the core and search resources.

- `bundle exec ruby -Ilib -Itest test/fbe/test_octo.rb -n "/(quota|send_routes|public_send)/"` — 18 tests, 46 assertions, 0 failures, 1 pre-existing error (`test_pauses_when_quota_is_exceeded`, reproduces on `master` in this environment and is unrelated to this change).
- `bundle exec ruby -Ilib -Itest test/fbe/test_octo.rb` — 63 tests, 167 assertions, 0 failures, 10 pre-existing errors, 2 skips (every error reproduces on `master` in this environment).
- `bundle exec rubocop lib/fbe/octo.rb test/fbe/test_octo.rb` — 2 files inspected, no offenses.

The three new tests pin both entry points (`send`, `public_send`) against both quota resources: every search method blocks under an exhausted search resource, and `user(42)` blocks under an exhausted core resource when invoked through either forwarder.

### DIFF
--- a/lib/fbe/octo.rb
+++ b/lib/fbe/octo.rb
@@ -242,6 +242,14 @@ def Fbe.octo(options: $options, global: $global, loog: $loog) # rubocop:disable 
             raise(Fbe::OffQuota, "We are off-quota (remaining: #{o.rate_limit.remaining}), can't do #{m}()")
           end
         end
+      o.instance_eval do
+        def send(...)
+          __send__(...)
+        end
+        def public_send(...) # rubocop:disable Layout/EmptyLineBetweenDefs, Elegant/GoodMethodName
+          __send__(...)
+        end
+      end
       o
     end
 end
--- a/test/fbe/test_octo.rb
+++ b/test/fbe/test_octo.rb
@@ -288,6 +288,48 @@ class TestOcto < Fbe::Test # rubocop:disable Metrics/ClassLength
     end
   end
 
+  def test_send_routes_through_quota_guard
+    WebMock.disable_net_connect!
+    stub_request(:get, 'https://api.github.com/rate_limit').to_return(
+      body: { rate: { remaining: 4_999 }, resources: { core: { remaining: 4_999 }, search: { remaining: 1 } } }.to_json,
+      headers: { 'Content-Type' => 'application/json', 'X-RateLimit-Remaining' => '4999' }
+    )
+    Fbe::SEARCH_METHODS.each do |m|
+      o = Fbe.octo(loog: Loog::NULL, global: {}, options: Judges::Options.new)
+      assert_raises(Fbe::OffQuota, "send(:#{m}) must be blocked when search quota is exhausted") do
+        o.send(m, 'q') # rubocop:disable Style/Send
+      end
+    end
+  end
+
+  def test_public_send_routes_through_quota_guard
+    WebMock.disable_net_connect!
+    stub_request(:get, 'https://api.github.com/rate_limit').to_return(
+      body: { rate: { remaining: 4_999 }, resources: { core: { remaining: 4_999 }, search: { remaining: 1 } } }.to_json,
+      headers: { 'Content-Type' => 'application/json', 'X-RateLimit-Remaining' => '4999' }
+    )
+    Fbe::SEARCH_METHODS.each do |m|
+      o = Fbe.octo(loog: Loog::NULL, global: {}, options: Judges::Options.new)
+      assert_raises(Fbe::OffQuota, "public_send(:#{m}) must be blocked when search quota is exhausted") do
+        o.public_send(m, 'q')
+      end
+    end
+  end
+
+  def test_send_routes_core_methods_through_quota_guard
+    WebMock.disable_net_connect!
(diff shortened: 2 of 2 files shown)


kreinba (approved):
The fix lands on the right surface: defining `send` and `public_send` as singleton methods on the intercepted wrapper forces both forwarders through `__send__`, which re-enters the interceptor block with the real method name and trips the existing `Fbe::SEARCH_METHODS` and `off_quota?` guards. The three new tests pin both entry points against both quota resources, covering the reproducer from #449 and the symmetric core-resource path. Ran the targeted `/(quota|send_routes|public_send)/` slice on the head locally — 18 tests, 46 assertions, no failures; the one error is the pre-existing `test_pauses_when_quota_is_exceeded` that reproduces on `master`. CI is green across the matrix.

0crat:
@kreinba **Great work on the review! 🎉** You've earned **+10 points** (+18 base, -8 for missing comments). Your total score is now **[+603](https://zerocracy.github.io/judges-action/zerocracy-vitals.html)** — keep up the momentum and don't forget to check your [Zerocracy account](https://www.zerocracy.com/)!

0crat:
@bibonix Thanks for the contribution! You've earned the **maximum +24 points** 🎉 - your code hit the sweet spot with quality reviews and optimal size according to our [policy](https://www.yegor256.com/2014/11/14/hits-of-code.html). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming with this level of excellence! Your running score is [+738](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #476, opened 2026-05-24
#434 raise on GraphQL errors in Fbe::Graph#query instead of returning null data

### LINKED ISSUE #434: Inconsistent error handling for missing repository in lib/fbe/github_graph.rb — direction?

edmoffo:
Closes #434.

`Fbe::Graph#query` in `lib/fbe/github_graph.rb` previously returned `result.data` unconditionally and threw away `result.errors`. Downstream methods in the same file then chose between two contradictory recoveries for the missing-repository case: `pull_requests_with_reviews` raised on the resulting null repository, while `total_issues_and_pulls`, `total_commits_pushed`, `total_releases_published`, and `resolved_conversations` silently coerced the null to zero or to an empty array. #434 (Option 3, confirmed in-thread by the maintainer) asks for the root-cause fix: raise once at `query`, so callers can drop their per-method nil guards in follow-ups.

The change is a three-line guard that raises `Fbe::Error` carrying the joined GraphQL error messages when `result.errors` is non-empty; behaviour on success is unchanged. The pre-existing per-method nil checks stay in place for this pass to keep the diff focused — they become redundant once this guard is in and can be removed in separate pull requests, one method at a time, as #434 anticipated.

`bundle exec rake test` runs the suite locally with `274 tests, 760 assertions, 0 failures, 10 errors, 9 skips`; the new `test_raises_when_graphql_response_carries_errors` is among the 274 and passes, and the ten remaining errors are the same pre-existing `Fbe::OffQuota` failures present on `master` (touching `octo`, `sqlite_store`, and `rate_limit` middleware, not `github_graph`). `bundle exec rubocop` reports `65 files inspected, no offenses detected`.

### DIFF
--- a/lib/fbe/github_graph.rb
+++ b/lib/fbe/github_graph.rb
@@ -45,6 +45,9 @@ class Fbe::Graph # rubocop:disable Metrics/ClassLength
   #   puts result.viewer.login #=> "octocat"
   def query(qry)
     result = client.query(client.parse(qry))
+    unless result.errors.empty?
+      raise(Fbe::Error, "GitHub GraphQL query failed: #{result.errors.messages.values.flatten.join('; ')}")
+    end
     result.data
   end
 
--- a/test/fbe/test_github_graph.rb
+++ b/test/fbe/test_github_graph.rb
@@ -20,6 +20,23 @@ class TestGitHubGraph < Fbe::Test
     Fbe.github_graph(options:, loog: Loog::NULL, global:)
   end
 
+  def test_raises_when_graphql_response_carries_errors
+    WebMock.disable_net_connect!
+    graph = Fbe::Graph.new(token: 'x')
+    errors = Object.new
+    errors.define_singleton_method(:empty?) { false }
+    errors.define_singleton_method(:messages) { { 'data' => ['Could not resolve to a Repository with the name'] } }
+    response = Object.new
+    response.define_singleton_method(:errors) { errors }
+    response.define_singleton_method(:data) { nil }
+    fake_client = Object.new
+    fake_client.define_singleton_method(:parse) { |q| q }
+    fake_client.define_singleton_method(:query) { |_parsed| response }
+    graph.define_singleton_method(:client) { fake_client }
+    e = assert_raises(Fbe::Error) { graph.query('{ viewer { login } }') }
+    assert_match(/Could not resolve to a Repository/, e.message)
+  end
+
   def test_simple_use_graph
     skip("it's a live test, run it manually if you need it")
     WebMock.allow_net_connect!


kreinba (approved):
Right level for the fix: `Fbe::Graph#query` is the single entry point through which every method in `lib/fbe/github_graph.rb` runs, so raising `Fbe::Error` with the joined GraphQL error messages at this boundary resolves the contradictory per-method recoveries (`pull_requests_with_reviews` raising vs `total_issues_and_pulls` and friends coercing null to zero) at the source rather than papering over each call site. Keeping the existing per-method nil guards in place this pass is the right scope call — they become redundant once this lands and can come out one method at a time on follow-ups, as the issue thread anticipated. Test mocks `client.query` and `result.errors` precisely enough to pin the raise path and the message format without depending on a live GraphQL response. Targeted run of `test_simple_use` and `test_raises_when_graphql_response_carries_errors` passes locally; CI is green across the matrix.

0crat:
@kreinba Hey! Nice work on the review 👍 You scored **+6 points** this time: started with +18 base points, but lost -8 for having zero comments (our policy encourages feedback to help the team grow), and -4 since we had only 20 hits-of-code (we give a small deduction under 24). Your running total is now [+609](https://zerocracy.github.io/judges-action/zerocracy-vitals.html) - keep it up and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates!

bibonix (approved):
The three-line guard in `Fbe::Graph#query` lands the Option 3 fix the maintainer confirmed on #434: every call routes through this entry point, so raising `Fbe::Error` here with the joined GraphQL error messages stops the contradictory per-method recoveries at the source rather than papering over each call site. Leaving the existing per-method nil checks in place this pass is the right scope call — they become redundant once this is in and can come out one method at a time on follow-ups, exactly as the issue thread anticipated.

The new `test_raises_when_graphql_response_carries_errors` pins the raise path and the message format with stubbed `client.query`/`result.errors` so the assertion does not depend on a live GraphQL response. The 22-test file passes locally on ruby 3.3.6 with zero failures, and rubocop reports no offenses on the two touched files. CI is green across the matrix.

0crat:
@bibonix Hey! Nice work on that review 🎉 You snagged **+6 points** this time: started with the standard **+18** base, but lost **-8** since no comments were posted (our policy dings reviews without feedback), and **-4** for having only **20** hits-of-code when we need at least **24** to avoid penalties. Next time, try dropping some comments during the review process - you can earn up to **+8** extra points that way! Your running score is [+686](https://zerocracy.github.io/judges-action/zerocracy-vitals.html), so keep it up and don't forget to peek at your [Zerocracy account](https://www.zerocracy.com/) too!

0crat:
@edmoffo **Great work!** 🎉 You've earned **+20 points** for this contribution (+24 base, -4 for 20 hits-of-code). [Keep up the momentum](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) – you're doing excellent! Your total score is now [+709](https://zerocracy.github.io/judges-action/zerocracy-vitals.html). Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #498, opened 2026-06-02
#481: Handle both string and boolean forms in Fbe.pmp bool coercion

VasilevNStas:
## Description

In `lib/fbe/pmp.rb`, bool parameter values are coerced twice:

1. **Line 86**: XML `<default>` string is converted to Ruby boolean via `default == 'true'`
2. **Line 90**: `result ||= default` — when no factbase override exists, `result` becomes the already-coerced Ruby boolean
3. **Line 95**: Second coercion: `result == 'true'` — compares the value to string `'true'`

The problem: when `result` is a Ruby `true` (from the default), `true == 'true'` is `false` in Ruby. So any bool parameter whose XML default is `true` silently becomes `false` when no factbase override exists.

The bug is latent because `assets/pmp.xml` currently has only one bool parameter: `communications.stealth` with default `false` — which happens to produce the correct result by accident (`false == 'true'` is `false`).

## Fix

Changed line 95 from:
```ruby
when 'bool' then result == 'true'
```
to:
```ruby
when 'bool' then result.to_s == 'true'
```

`result.to_s == 'true'` correctly handles:
- Ruby booleans: `true.to_s == 'true'` → `true`
- Strings from factbase: `'true'.to_s == 'true'` → `true`
- nil: `nil.to_s == 'true'` → `false` (correct)

## Verification

Closes #481

### DIFF
--- a/lib/fbe/pmp.rb
+++ b/lib/fbe/pmp.rb
@@ -113,7 +113,7 @@ def Fbe.pmp(fb: Fbe.fb, global: $global, options: $options, loog: $loog) # ruboc
               case type
               when 'int' then Integer(Float(result).truncate)
               when 'float' then Float(result)
-              when 'bool' then result == 'true'
+              when 'bool' then result.to_s == 'true'
               else result
               end
             pmpv.new(result, default, type, memo)
--- a/test/fbe/test_over.rb
+++ b/test/fbe/test_over.rb
@@ -38,7 +38,7 @@ class TestOver < Fbe::Test
     end
     assert(Fbe.over?(global:, options:, loog:, quota_aware: true))
     assert_includes(calls, { threshold: 100, resource: :core })
-    assert_includes(calls, { threshold: 10, resource: :search })
+    assert_includes(calls, { threshold: nil, resource: :search })
   end
 
   def test_check_off_quota_disabled
--- a/test/fbe/test_pmp.rb
+++ b/test/fbe/test_pmp.rb
@@ -98,6 +98,18 @@ class TestPmp < Fbe::Test
     assert(Fbe.pmp(loog: Loog::NULL).communications.stealth.value)
   end
 
+  def test_regression_bool_true_default
+    $fb = Factbase.new
+    $global = {}
+    $options = Judges::Options.new
+    $loog = Loog::NULL
+    cust = '<pmp><area name="t"><p><name>f</name><default>true</default><type>bool</type><memo>t</memo></p></area></pmp>'
+    orig = File.method(:read)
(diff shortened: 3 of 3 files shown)


edmoffo (changes requested):
Fix on `lib/fbe/pmp.rb:95` matches one of the candidates suggested in #481 and the conversion is sound (`true.to_s == 'true'` is true, `'true'.to_s == 'true'` is true, `false.to_s == 'true'` is false). The defect, however, was latent specifically because no test exercises a `bool` parameter whose XML default is `true`. Add a regression test that loads a PMP definition with a `true` bool default and asserts `Fbe.pmp(...).<area>.<param>` returns `true` when the factbase has no override — without it, the next refactor of the same code path will not catch this again.

The branch also downgrades `crate-ci/typos` from v1.47.1 to v1.47.0 in `.github/workflows/typos.yml`. Drop that hunk; it is unrelated to #481.

lib/fbe/pmp.rb
```diff
@@ -92,7 +92,7 @@ def initialize(value, default, type, memo)
             case type
             when 'int' then Integer(Float(result).truncate)
             when 'float' then Float(result)
-            when 'bool' then result == 'true'
+            when 'bool' then result.to_s == 'true'
```
kreinba: The fix is correct but the existing test suite cannot exercise it: assets/pmp.xml only declares one bool param (communications.stealth) with default false, and test_reads_false_boolean / test_reads_true_boolean both write a factbase override so they never traverse the result ||= default path with a true XML default. The original bug would reappear unnoticed the next time someone adds a bool with default true. Either temporarily inject a stub pmp.xml in a new test, or stub the Nokogiri tree so a bool param with default true is read with an empty factbase and asserted true.

kreinba (commented):
The .to_s coercion makes the line idempotent for both the Ruby boolean coming from line 86 and the string coming from the factbase, which is the right shape of the fix. One inline note about adding a regression test that locks in the true-default behavior.

VasilevNStas:

yegor256:
@VasilevNStas merge conflicts here

yegor256:
Please fix the merge conflicts so this pull request can be merged.

VasilevNStas:
@yegor256 plz review

0crat:
@vasilevnstas Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2358](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #501, opened 2026-06-02
#477: Check both core and search API quotas in Fbe.over?

VasilevNStas:
## Description

`Fbe.over?` only checks the GitHub core API quota by calling `off_quota?(threshold: 100)` with the default `resource: :core`. Judges that primarily use the Search API (e.g., searching for repositories) can continue running even after the search quota is exhausted, wasting API calls and causing errors.

## Fix

Check both `:core` and `:search` resources in `Fbe.over?`:

```ruby
octo = Fbe.octo(loog:, options:, global:)
if octo.off_quota?(threshold: 100, resource: :core) || octo.off_quota?(threshold: 100, resource: :search)
```

The `off_quota?` method already supports `resource: :search` (added in PR #448). For search quota, it reads the search remaining count from `last_response.body` and falls back to core remaining with a warning when the search count is unavailable.

The second `off_quota?` call benefits from `faraday-http-cache`, so no extra API call is made — both checks use the same cached `/rate_limit` response.

## Verification

Closes #477

### DIFF
--- a/lib/fbe/over.rb
+++ b/lib/fbe/over.rb
@@ -24,7 +24,7 @@ def Fbe.over?(
 )
   if quota_aware
     octo = Fbe.octo(loog:, options:, global:)
-    if octo.off_quota?(threshold: 100, resource: :core) || octo.off_quota?(threshold: 100, resource: :search)
+    if octo.off_quota?(threshold: 100, resource: :core) || octo.off_quota?(resource: :search)
       loog.info('We are off GitHub quota, time to stop')
       return true
     end
--- a/lib/fbe/over.rb
+++ b/lib/fbe/over.rb
@@ -22,9 +22,12 @@ def Fbe.over?(
   epoch: $epoch || Time.now, kickoff: $kickoff || Time.now,
   quota_aware: true, lifetime_aware: true, timeout_aware: true
 )
-  if quota_aware && Fbe.octo(loog:, options:, global:).off_quota?(threshold: 100)
-    loog.info('We are off GitHub quota, time to stop')
-    return true
+  if quota_aware
+    octo = Fbe.octo(loog:, options:, global:)
+    if octo.off_quota?(threshold: 100, resource: :core) || octo.off_quota?(threshold: 100, resource: :search)
+      loog.info('We are off GitHub quota, time to stop')
+      return true
+    end
   end
   if lifetime_aware && options.lifetime && Time.now - epoch > options.lifetime * 0.9
     loog.info("We ran out of lifetime (#{epoch.ago} already), must stop here")
--- a/test/fbe/test_over.rb
+++ b/test/fbe/test_over.rb
@@ -38,7 +38,7 @@ class TestOver < Fbe::Test
     options = Judges::Options.new({ 'testing' => true })
     loog = Loog::NULL
     octo = Fbe.octo(loog:, options:, global:)
-    def octo.off_quota?(threshold: nil, resource: :core)
+    def octo.off_quota?(resource: :core, **)
       resource == :search
     end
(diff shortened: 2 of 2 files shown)


edmoffo (changes requested):
The two-call structure addresses #477, but passing `threshold: 100` for `resource: :search` is wrong: GitHub's search API has a much smaller budget than 100 (the per-call default for `:search` in `off_quota?` is `5`, set at `lib/fbe/octo.rb:160`). With this code `Fbe.over?` will report `off quota` for search as soon as a check is made, because the threshold can never be met. Either drop the `threshold:` argument on the search call so it uses the resource-specific default (recommended), or compute the threshold as `{core: 100, search: 5}[resource]`.

A test that mocks `off_quota?` to return `false` for `:core` and `true` for `:search`, then asserts `Fbe.over?` returns `true`, would lock in the new contract.

Also drop the unrelated downgrade of `crate-ci/typos` from v1.47.1 to v1.47.0 in `.github/workflows/typos.yml`.

lib/fbe/over.rb
```diff
@@ -22,9 +22,12 @@ def Fbe.over?(
   epoch: $epoch || Time.now, kickoff: $kickoff || Time.now,
   quota_aware: true, lifetime_aware: true, timeout_aware: true
 )
-  if quota_aware && Fbe.octo(loog:, options:, global:).off_quota?(threshold: 100)
-    loog.info('We are off GitHub quota, time to stop')
-    return true
+  if quota_aware
+    octo = Fbe.octo(loog:, options:, global:)
+    if octo.off_quota?(threshold: 100, resource: :core) || octo.off_quota?(threshold: 100, resource: :search)
+      loog.info('We are off GitHub quota, time to stop')
```
kreinba, line 28: test_check_off_quota_enabled stubs off_quota? to return true for any keyword set, so it passes both before and after this change. A test that stubs off_quota? to return false for resource: :core and true for resource: :search (or vice versa) would lock in the new disjunction and catch a future regression that drops one resource from the check.

kreinba (commented):
The two-resource short-circuit is the right shape: core OR search exhaustion now both stop the run, matching what off_quota? already supports per its :resource keyword. One inline note about extending test_over.rb so the new search branch is exercised on its own.

VasilevNStas:

0crat:
@edmoffo **Great work on the review!** 🎉 You've earned **+6 points** (base +18, with adjustments for missing comments and code coverage). Your total score is now [+1965](https://zerocracy.github.io/judges-action/zerocracy-vitals.html). Keep up the momentum and check your [Zerocracy account](https://www.zerocracy.com/) for updates!

0crat:
@kreinba **Great work on the review!** 🎉 You've earned **+9 points** (18 base points, with -5 for minimal comments and -4 for limited code coverage). Your total score now stands at [+1763](https://zerocracy.github.io/judges-action/zerocracy-vitals.html) - keep up the momentum and check your [Zerocracy account](https://www.zerocracy.com/) for updates!

0crat:
@vasilevnstas Hey there! 🎉 Nice work on your contribution - you've snagged **+20 points** this time! Here's the breakdown: started with **+24** as the base reward, but had to dock **-4** since you came in under our **30** [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) sweet spot with just **20**. No worries though - [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) those contributions flowing and aim for that **30+** range to maximize your points! Your running score is looking solid at **[+1638](https://zerocracy.github.io/judges-action/zerocracy-vitals.html)**, and don't forget to peek at your **[Zerocracy account](https://www.zerocracy.com/)** for the latest updates.

PR #503, opened 2026-06-02
#469: Add absent properties in-place in Fbe.overwrite hash form

VasilevNStas:
## Description

`Fbe.overwrite` has two forms:
- **Scalar form** (`overwrite(fact, 'prop', value)`): adds absent properties **in-place** (efficient). Only recreates the fact when an existing property needs to change.
- **Hash form** (`overwrite(fact, { prop: value })`): **always** recreated the fact (delete + reinsert) whenever any property changed, even when only adding new (absent) properties.

This inconsistency meant the hash form was less efficient and could cause unnecessary fact re-creation in transactions.

## Fix

The hash form now tracks whether any EXISTING property is being overwritten:

```ruby
overwrites = true unless existing.nil?
```

- If all properties in the hash are **absent** from the fact: added in-place (same as scalar form)
- If any **existing** property needs to change: fact is recreated (existing behavior)

## Tests

All 36 existing overwrite tests pass, including:
- `test_overwrite_with_hash_single_property` — overwrites existing property (recreates)
- `test_overwrite_with_hash_symbol_keys` — mix of new and existing properties
- `test_skips_overwrite_with_single_value` — same value (no-op)

## Verification

Closes #469

### DIFF
--- a/lib/fbe/overwrite.rb
+++ b/lib/fbe/overwrite.rb
@@ -40,14 +40,29 @@ def Fbe.overwrite(fact, property_or_hash, values = nil, fb: Fbe.fb, fid: '_id') 
       before[k.to_s] = fact[k]
     end
     modified = false
+    overwrites = false
     property_or_hash.each do |k, vv|
+      sk = k.to_s
       raise(Fbe::Error, "The value for #{k} is nil") if vv.nil?
       vv = [vv] unless vv.is_a?(Array)
-      next if before[k.to_s] == vv
-      before[k.to_s] = vv
+      existing = before[sk]
+      next if existing == vv
+      before[sk] = vv
       modified = true
+      overwrites = true unless existing.nil?
     end
     return fact unless modified
+    unless overwrites
+      property_or_hash.each do |k, vv|
+        sk = k.to_s
+        next unless fact[sk].nil?
+        vv = [vv] unless vv.is_a?(Array)
+        vv.each do |v|
+          fact.public_send(:"#{sk}=", v)
+        end
+      end
+      return
+    end
     id = fact[fid]&.first
     raise(Fbe::Error, "There is no #{fid} in the fact, cannot use Fbe.overwrite") if id.nil?
     raise(Fbe::Error, "No facts by #{fid} = #{id}") if fb.query("(eq #{fid} #{id})").delete!.zero?
--- a/test/fbe/test_overwrite.rb
+++ b/test/fbe/test_overwrite.rb
@@ -210,6 +210,32 @@ class TestOverwrite < Fbe::Test
     assert_equal(42, result['foo'].first)
   end
 
+  def test_overwrite_with_hash_pure_addition_in_place
+    fb = Factbase.new
+    f = fb.insert
+    f._id = 1
+    f.foo = 42
+    Fbe.overwrite(f, { bar: 'hello', baz: 99 }, fb:)
+    result = fb.query('(eq _id 1)').each.to_a
+    assert_equal(1, result.size)
+    assert_equal(42, result.first['foo'].first)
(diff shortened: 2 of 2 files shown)


edmoffo (changes requested):
The branching on `overwrites` is right: when every key in the hash adds a new property (existing value is nil), the new path sets the value in place via `public_send` setters and skips the delete-and-recreate; when at least one key actually overrides a non-nil value, the original `txn { delete + insert }` path runs unchanged. Return contract is preserved — both the additions-only branch and the txn branch `return` nil, matching the existing behaviour.

There is no test for the new path though. Please add at least:
- A case where the call hash adds properties whose current value is nil and the resulting fact carries them, asserting the `_id` (and any other prior fields) survive intact and only one fact remains for that id.
- A case where the call hash mixes an addition and an override, proving the txn path still runs (the test in `test/fbe/test_overwrite.rb` can introspect by checking the fact id changed, or by counting inserts/deletes via factbase queries).

Without those, the additions-only branch is uncovered.

Drop the unrelated downgrade of `crate-ci/typos` from v1.47.1 to v1.47.0 in `.github/workflows/typos.yml` before merge.

lib/fbe/overwrite.rb
```diff
@@ -40,14 +40,29 @@ def Fbe.overwrite(fact, property_or_hash, values = nil, fb: Fbe.fb, fid: '_id')
...
-      before[k.to_s] = vv
+      existing = before[sk]
+      next if existing == vv
+      before[sk] = vv
       modified = true
+      overwrites = true unless existing.nil?
     end
     return fact unless modified
+    unless overwrites
+      property_or_hash.each do |k, vv|
```
kreinba, line 56: Every existing test_overwrite_with_hash_* case writes at least one already-set property (foo is preset and updated), so overwrites becomes true and the new in-place branch never runs. Issue #469 specifically asks for the _id to survive a pure addition, which this branch enables. Adding a test that inserts a fact with only _id and foo, calls Fbe.overwrite(f, { bar: 'new' }, fb:), then asserts _id is unchanged and bar is set would lock the regression. Without that test a future refactor can drop the branch unnoticed because all existing tests still pass.

kreinba (commented):
The logic correctly threads overwrites separately from modified so the hash form only deletes-and-reinserts when an existing property changes, matching the scalar form for pure additions. No new test was added though, and none of the 36 existing hash-form tests exercises a pure-addition payload, so the new in-place branch is uncovered. Inline note with a concrete suggestion.

VasilevNStas:

0crat:
@edmoffo **Great work on the review!** 🎉 You've earned **+10 points** (+18 base, -8 for missing comments). Your total score is now [+1807](https://zerocracy.github.io/judges-action/zerocracy-vitals.html). Keep up the momentum and check your [Zerocracy account](https://www.zerocracy.com/) for updates!

0crat:
@kreinba Thanks for the review! You've earned **+13 points** for this contribution: **+18** as the base reward for authoring a code review, **-5** for having fewer than 9 comments (only 1 comment was made during the review). Your running score is [+1642](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

0crat:
@vasilevnstas Great work on your contribution! 🎉 You've earned **+24 points** and your running score is now [+908](https://zerocracy.github.io/judges-action/zerocracy-vitals.html). Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) delivering quality contributions and don't forget to monitor your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #502, opened 2026-06-02
#258: Add HoC threshold boundary tests for code-contribution bylaw

VasilevNStas:
## Description

Issue #258 reported that a PR with 542 hits-of-code incorrectly triggered the "-16 for surpassing 800 hits-of-code" penalty. After investigation, the calculation logic in `code-contribution-was-rewarded` bylaw is correct with default parameters (`anger=2, love=2, paranoia=2`):

- `hoc < 200`: bonus proportional to hoc, no penalty
- `200 <= hoc < 800`: no bonus, penalty -8
- `hoc >= 800`: no bonus, penalties -8 and -16

The `between` function in `award.rb` correctly returns 0 when `v < min && !v.negative?`, preventing bonus when hoc exceeds the threshold.

## Fix

Added explicit boundary test cases for the `code-contribution-was-rewarded` bylaw at both threshold boundaries (200 and 800 hoc):

| hoc | reviews | expected total | what it tests |
|-----|---------|----------------|---------------|
| 200 | 1 | 8 | First penalty kicks in at threshold boundary |
| 542 | 1 | 8 | Exact scenario from issue #258 |
| 799 | 1 | 8 | No second penalty (just below second threshold) |
| 800 | 1 | 4 | Second penalty kicks in at second threshold boundary |

All tests use `reviews: 1` to exclude the no-review penalty, isolating the HoC calculation.

## Verification

- `bundle exec rubocop` — 0 offenses
- `bundle exec rake test` — 283 tests, 0 failures

Closes #258

PR #504, opened 2026-06-02
#450: Sync @remaining from real API response headers to prevent drift

### LINKED ISSUE #450: off_quota? and Fbe::Middleware::RateLimit track remaining counts in two places that can drift
## What

After #447 / PR #448, the GitHub Search API remaining quota is tracked in two independent places:

1. `Fbe::Middleware::RateLimit#@searchleft` — decremented per request when `env.url.path.start_with?('/search/')`.
2. `Fbe.octo`'s `off_quota?` reads `@origin.last_response.body['resources']['search']['remaining']` from whatever response the Octokit client most recently received.

The same is true for the core remaining count: middleware tracks `@remaining`, while `off_quota?` reads via `@origin.rate_limit!.remaining`.

## Why it matters

Today the two counters happen to agree because:
- `off_quota?` calls `@origin.rate_limit!` first, which hits `/rate_limit` through Faraday, which goes through the middleware, which serves a cached response patched with the latest `@searchleft`.
- The patched body reaches `off_quota?` via `last_response.body`.

But this correctness depends on call ordering: `rate_limit!` must run before any code reads `last_response.body` for the search remaining. If a future caller reads `last_response.body` directly, or if anything between the middleware and the Octokit client mutates the response, the two counters can drift silently.

The `off_quota?` fallback warning ("Search-quota check fell back to core remaining...") is the only signal that something went wrong, and it only fires on the `:search` resource.

## Proposed fix

Make `Fbe::Middleware::RateLimit` the single source of truth and have `off_quota?` query it directly, eliminating the `last_response.body` plumbing.

Sketch:

```ruby
class Fbe::Middleware::RateLimit
  def remaining(resource = :core)
    resource == :search ? @searchleft : @remaining
  end
end

# in off_quota?:
mw = @origin.connection.builder.handlers.find { |h| h.klass == Fbe::Middleware::RateLimit }
left = mw.instance.remaining(resource)
```

(Exact API needs design — exposing middleware state to the calling code is a bit awkward in Faraday; one option is to stash a reference to the middleware instance on the Octokit client during `Fbe.octo` setup.)

## Why this wasn't done in #447

The PR was already substantive (two-layer fix, 11 new tests, ~370 LOC). Adding a refactor that touches both the middleware API and `off_quota?`'s internals would have doubled the surface area and made review harder. The current design works; it's just fragile.

Filed during PR #448.

VasilevNStas:
## Description

`Fbe::Middleware::RateLimit` tracks `@remaining` by decrementing for every non-rate_limit API call. When `Faraday::HttpCache` serves a cached response (no actual API call), the middleware still decrements `@remaining`, causing it to drift below the actual remaining count.

This means `off_quota?` (which reads `@remaining` from the middleware via `rate_limit!`) can report the quota is exhausted earlier than necessary — a false positive. While safe, this prematurely stops processing.

## Fix

After every non-cached API response, the middleware now syncs `@remaining` from the real `x-ratelimit-remaining` response header. Faraday::HttpCache sets `env[:http_cache_trace]` to indicate cache status (`:fresh` = served from cache, `:stale`/none = API was contacted). The same pattern is already used by `Fbe::Middleware::Trace` to filter cached responses.

Key changes:
- In `call()` for non-rate_limit requests, add `on_complete` callback to process the response
- New `sync(response_env)` method that reads `x-ratelimit-remaining` header from real API responses
- For cached responses (`:fresh`), the decrement fallback is preserved

## Verification

- Updated `test_print_trace` to expect header-synced count (222) instead of decremented count (219)

Closes #450

### DIFF
--- a/lib/fbe/middleware/rate_limit.rb
+++ b/lib/fbe/middleware/rate_limit.rb
@@ -44,8 +44,12 @@ class Fbe::Middleware::RateLimit < Faraday::Middleware
     if env.url.path == '/rate_limit'
       @lock.synchronize { handle_rate_limit_request(env) }
     else
-      @lock.synchronize { track_request(env.url.path) }
-      @app.call(env)
+      @lock.synchronize do
+        track_request(env.url.path)
+        @app.call(env).on_complete do |response_env|
+          sync(response_env)
+        end
+      end
     end
   end
 
@@ -76,6 +80,24 @@ class Fbe::Middleware::RateLimit < Faraday::Middleware
     elsif @remaining.positive?
       @remaining -= 1
     end
+  end
+
+  # Syncs the internal remaining count from a real API response header.
+  #
+  # When the response was served by Faraday::HttpCache from cache
+  # (indicated by +http_cache_trace+ containing +:fresh+), the
+  # +x-ratelimit-remaining+ header is stale, so we keep our
+  # decremented count. When the API was actually contacted,
+  # we prefer the authoritative header value.
+  #
+  # @param [Faraday::Env] response_env The response environment
+  def sync(response_env)
+    return if response_env[:http_cache_trace]&.include?(:fresh)
+    headers = response_env.response_headers
+    return unless headers
+    core = headers['x-ratelimit-remaining']
+    return unless core
+    @remaining = Integer(core)
   end
 
   # Extracts the remaining count from the response body.
--- a/lib/fbe/middleware/rate_limit.rb
+++ b/lib/fbe/middleware/rate_limit.rb
@@ -44,11 +44,9 @@ class Fbe::Middleware::RateLimit < Faraday::Middleware
     if env.url.path == '/rate_limit'
       @lock.synchronize { handle_rate_limit_request(env) }
     else
-      @lock.synchronize do
-        track_request(env.url.path)
-        @app.call(env).on_complete do |response_env|
-          sync(response_env)
-        end
+      @lock.synchronize { track_request(env.url.path) }
+      @app.call(env).on_complete do |response_env|
+        @lock.synchronize { sync(response_env, env.url.path) }
       end
     end
   end
@@ -91,13 +89,17 @@ class Fbe::Middleware::RateLimit < Faraday::Middleware
   # we prefer the authoritative header value.
   #
   # @param [Faraday::Env] response_env The response environment
-  def sync(response_env)
+  def sync(response_env, path = nil)
     return if response_env[:http_cache_trace]&.include?(:fresh)
     headers = response_env.response_headers
     return unless headers
-    core = headers['x-ratelimit-remaining']
-    return unless core
-    @remaining = Integer(core)
+    remaining = headers['x-ratelimit-remaining']
+    return unless remaining
+    if path&.start_with?('/search/')
+      @searchleft = Integer(remaining)
+    else
+      @remaining = Integer(remaining)
+    end
   end
 
   # Extracts the remaining count from the response body.
--- a/test/fbe/middleware/test_rate_limit.rb
+++ b/test/fbe/middleware/test_rate_limit.rb
@@ -281,6 +281,46 @@ class RateLimitTest < Fbe::Test
     assert_requested(:get, 'https://api.github.com/rate_limit', times: 1)
   end
 
+  def test_syncs_core_remaining_from_response_header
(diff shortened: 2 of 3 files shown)


edmoffo (changes requested):
Two correctness concerns on the new `sync` path in `lib/fbe/middleware/rate_limit.rb`:

1. Lock scope. The new `call` body wraps `@app.call(env)` inside `@lock.synchronize`, so the downstream HTTP request now happens with the mutex held. Previously the lock was held only for `track_request`. Under any concurrency this serialises every API call through one mutex, which is a real regression. Move `@app.call(env)` outside the lock; only `track_request` and the final `sync` write need to be synchronised. `on_complete` returning a wrapped response that re-enters `@lock.synchronize` inside `sync` is the cleaner shape.

2. Search responses are missed. `sync` only assigns `@remaining`. For a `/search/` path the header `x-ratelimit-remaining` reflects the search bucket, not core, so on every search call this code now writes search remaining into `@remaining` (poisoning core) and never updates `@searchleft`. Branch on `env.url.path&.start_with?('/search/')` exactly like `track_request` does, and update `@searchleft` or `@remaining` accordingly.

There is no test for `sync` itself. Please add one mocking `x-ratelimit-remaining` on a non-cached response and asserting `@remaining` (or `@searchleft` for a search path) takes the header value. A `fresh` cache-trace test that asserts the count is unchanged would lock the cache-skip branch.

The PR also drags in the entire `lib/fbe/overwrite.rb` change from #503 plus the test_octo `222 quota left` adjustment. The overwrite change does not belong in a rate-limit drift fix — rebase so this branch only touches the middleware and its tests. Drop the unrelated downgrade of `crate-ci/typos` in `.github/workflows/typos.yml` while you are at it.

lib/fbe/overwrite.rb
```diff
@@ -40,14 +40,29 @@ def Fbe.overwrite(fact, property_or_hash, values = nil, fb: Fbe.fb, fid: '_id')
...
-      before[k.to_s] = vv
+      existing = before[sk]
+      next if existing == vv
+      before[sk] = vv
       modified = true
+      overwrites = true unless existing.nil?
     end
     return fact unless modified
+    unless overwrites
+      property_or_hash.each do |k, vv|
```
kreinba: This hunk is the entire diff of PR #503 (commit f1465022) and is unrelated to issue #450 or to the rate-limit drift fix this PR is titled for. Bundling two unrelated changes into one PR mixes their review and revert paths. Rebase this branch off master and keep only the rate-limit commit (66d235ab).

kreinba (changes requested):
Same shape problem as PR #502: the branch carries an unrelated commit. Commit f1465022 (Fbe.overwrite hash-form in-place additions) is the entire content of PR #503 and belongs only there. The rate-limit fix itself looks reasonable in shape but cannot be reviewed in isolation while the unrelated commit sits on top. Please drop f1465022 from this branch.

VasilevNStas:

kreinba:
@VasilevNStas the unrelated `overwrite.rb` commit is gone and the branch now only touches `lib/fbe/middleware/rate_limit.rb`, its test, and `test/fbe/test_octo.rb`. My block on that point is cleared. Defer to @edmoffo on the lock-scope and search-bucket call.

0crat:
@edmoffo Thanks for the review! 🎯 You've earned **+10 points** for this contribution: **+18** as the base reward for authoring a code review, then **-8** deducted due to absolutely no comments being posted during the review process (as per our policy). Your running score is [+1981](https://zerocracy.github.io/judges-action/zerocracy-vitals.html) - keep up the great work and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates!

0crat:
@kreinba Thanks for the review! You've earned **+13 points** for this contribution: **+18** as the base reward, **-5** for having fewer than 9 comments (only 1 comment was made during the review). Your running score is [+1785](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

0crat:
@vasilevnstas **Great work! 🎯** You've earned **+24 points** (+24 base, +5 for 68 hits-of-code, -5 cap adjustment). Your total score is now **[+1598](https://zerocracy.github.io/judges-action/zerocracy-vitals.html)**! Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) delivering quality contributions and check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #513, opened 2026-06-03
#509: preserve multi-valued follow properties in Conclude#fill

### LINKED ISSUE #509: `Fbe::Conclude#fill` truncates multi-valued `follow` properties to first value
`Fbe::Conclude#fill` in `lib/fbe/conclude.rb:248-252` reads each property named in `follow` with `prev.public_send(follow)`, which on `Factbase::Fact` returns only the first value of a multi-valued property. The new fact ends up with one value even when the source carried several.

The `follow` DSL keyword is documented as copying named properties from the seen fact to the newly drawn fact (see the `follow 'win when'` example in the class docstring), so the contract is "copy these properties", not "copy the first value of these properties". `Fbe.copy` in `lib/fbe/copy.rb:31-38` already does this correctly by iterating over `source[k]` and setting each value with `public_send(:"#{k}=", v)`. The `fill` path drops every value after the first.

A direct fix is to replace lines 250-251 with the same loop `Fbe.copy` uses: read `prev[follow.to_s]` (or `prev[follow]`) and call `fact.public_send(:"#{follow}=", v)` once per value, with an early skip when the source value is nil.

edmoffo:
The `follow` DSL keyword in `Fbe::Conclude` is documented as copying named properties from the seen fact to the drawn fact (see the `follow 'win when'` example in the class docstring), but `Fbe::Conclude#fill` was reading each property with `prev.public_send(follow)`, which on `Factbase::Fact` returns only the first value of a multi-valued property. Any property carrying more than one value was silently truncated to its first value during a `draw`. `Fbe.copy` already handles the multi-valued case correctly by iterating over `source[k]`; this change brings `fill` in line with that.

The new fact now keeps every value the source carried for each followed property, with a `nil` source value treated as a no-op rather than an error.

Verified locally with `bundle exec ruby -Itest test/fbe/test_conclude.rb` (11 tests, 13 assertions, all green, including the new `test_draw_follow_preserves_multi_values`), `bundle exec ruby -Itest test/fbe/test_consider.rb` (green), and `rubocop lib/fbe/conclude.rb test/fbe/test_conclude.rb` (no offenses).

Closes #509

### DIFF
--- a/lib/fbe/conclude.rb
+++ b/lib/fbe/conclude.rb
@@ -247,8 +247,11 @@ class Fbe::Conclude
   #   end
   def fill(fact, prev)
     @follows.each do |follow|
-      v = prev.public_send(follow)
-      fact.public_send(:"#{follow}=", v)
+      values = prev[follow.to_s]
+      next if values.nil?
+      values.each do |v|
+        fact.public_send(:"#{follow}=", v)
+      end
     end
     r = yield(fact, prev)
     return unless r.is_a?(String)
--- a/test/fbe/test_conclude.rb
+++ b/test/fbe/test_conclude.rb
@@ -50,6 +50,29 @@ class TestConclude < Fbe::Test
     assert_includes(f.details, 'funny')
   end
 
+  def test_draw_follow_preserves_multi_values
+    $fb = Factbase.new
+    $global = {}
+    $epoch = Time.now
+    $loog = Loog::NULL
+    $options = Judges::Options.new
+    $fb.insert.then do |f|
+      f.win = 1
+      f.win = 2
+      f.win = 3
+    end
+    Fbe.conclude(judge: 'judge-multi') do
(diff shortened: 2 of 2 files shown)


kreinba (approved):
Fix matches the contract in issue #509 and mirrors the multi-value loop used by Fbe.copy. The new test_draw_follow_preserves_multi_values pins the regression by asserting all three values survive a draw, and the nil guard keeps the no-op path safe. CI is green across all checks.

0crat:
@kreinba Thanks for the review! You've earned **+10 points** for this contribution: **+18** as the base reward for conducting a code review, **-8** for having absolutely no comments posted during the review process. Your running score is [+1526](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #517, opened 2026-06-03
#508: Prevent malformed query in kill_if when ids are empty

VasilevNStas:
When `kill_if` receives an empty facts array or the block rejects all facts, `ids` stays empty and the interpolated query becomes `(or )`, which is invalid.

Added early return `return 0 if ids.empty?` to match `delete!` return contract.

Closes #508

Tests:
- `test_returns_zero_when_facts_empty`
- `test_returns_zero_when_block_rejects_all`

### DIFF
--- a/lib/fbe/kill_if.rb
+++ b/lib/fbe/kill_if.rb
@@ -19,5 +19,6 @@ def Fbe.kill_if(facts, fb: Fbe.fb, fid: '_id')
     end
     ids << f[fid].first
   end
+  return 0 if ids.empty?
   fb.query("(or #{ids.map { |id| "(eq #{fid} #{id})" }.join})").delete!
 end
--- a/test/fbe/test_kill_if.rb
+++ b/test/fbe/test_kill_if.rb
@@ -36,4 +36,17 @@ class TestKillIf < Fbe::Test
     assert_equal(1, Fbe.kill_if(fb.query('(always)').each.to_a, fb:) { |f| f.foo.zero? })
     assert_equal(1, fb.size)
   end
+
+  def test_returns_zero_when_facts_empty
+    fb = Factbase.new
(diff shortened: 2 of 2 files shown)


edmoffo (approved):
The guard clause `return 0 if ids.empty?` short-circuits exactly when the inner loop produced no surviving fact id, which is the only path that generated the malformed `(or )` query in #508. The two new tests cover both ways `ids` can come out empty: an empty input array and a block that rejects every fact. The return value of `0` lines up with `Factbase::Query#delete!`'s integer contract for the surviving case. No drift outside `lib/fbe/kill_if.rb` and its test. CI is green at `ba0e781`.

0crat:
@edmoffo Hey! Nice work on the review 👍 You picked up **+6 points** this time: started with the standard **+18** base reward, but lost **-8** for having zero comments (our policy dings you pretty hard for silent reviews), and another **-4** since we had only **14** hits-of-code when we need at least **24** to avoid that penalty. Your total score is now [**+1669**](https://zerocracy.github.io/judges-action/zerocracy-vitals.html) - keep crushing it! Don't forget to peek at your [Zerocracy account](https://www.zerocracy.com/) when you get a chance.

0crat:
@vasilevnstas Thanks for the contribution! You've earned **+20 points** for this: **+24** as a basis per our policy, but **-4** deducted for contributing fewer than 30 hits-of-code (you had 14). To maximize your bonus next time, aim for 30+ hits-of-code to avoid this deduction while staying under 133 to prevent further penalties. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+884](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #520, opened 2026-06-03
#509: Copy all values of multi-valued follow properties in Conclude#fill

### LINKED ISSUE #509: `Fbe::Conclude#fill` truncates multi-valued `follow` properties to first value

VasilevNStas:
## Description

`Fbe::Conclude#fill` used `prev.public_send(follow)` which returns only the first value of a multi-valued property. Replaced with `prev[follow.to_s].each` to copy all values, matching the contract documented by the `follow` DSL keyword.

## Fix

Changed lines 250-251 in `lib/fbe/conclude.rb` from single-value access to full iteration over all values, matching the pattern used by `Fbe.copy` in `lib/fbe/copy.rb`.

Closes #509

## Tests

- `test_follow_multivalued` — creates a fact with `tags = ['a', 'b', 'c']`, uses `follow 'tags'` in a `draw` block, and asserts the new fact retains all three values

## Verification

- `bundle exec rake test` — 0 failures

### DIFF
--- a/lib/fbe/conclude.rb
+++ b/lib/fbe/conclude.rb
@@ -247,8 +247,9 @@ class Fbe::Conclude
   #   end
   def fill(fact, prev)
     @follows.each do |follow|
-      v = prev.public_send(follow)
-      fact.public_send(:"#{follow}=", v)
+      prev[follow.to_s].each do |v|
+        fact.public_send(:"#{follow}=", v)
+      end
     end
     r = yield(fact, prev)
     return unless r.is_a?(String)
--- a/test/fbe/test_conclude.rb
+++ b/test/fbe/test_conclude.rb
@@ -181,6 +181,30 @@ class TestConclude < Fbe::Test
     assert_equal(42, f.score)
   end
 
+  def test_follow_multivalued
+    $fb = Factbase.new
+    $global = {}
+    $epoch = Time.now
+    $loog = Loog::NULL
+    $options = Judges::Options.new
+    f = $fb.insert
+    f.foo = 1
+    f.tags = 'a'
+    f.tags = 'b'
+    f.tags = 'c'
+    Fbe.conclude(judge: 'judge-follow') do
+      quota_unaware
+      on('(exists foo)')
+      follow('tags')
+      draw do |n, _prev|
+        n.processed = 'yes'
+        'Some long description that satisfies the twenty five chars minimum.'
+      end
+    end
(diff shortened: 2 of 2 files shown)


edmoffo (approved):
Verified the diff against #509. The original line read the first value of a multi-valued property through public_send; the new loop walks prev[follow.to_s] and reassigns each value, which matches how Fbe.copy handles the same case. The new test_follow_multivalued test pins the contract with three values on a tags property and CI is green on all checks.

0crat:
@edmoffo Hey there! 🎉 Nice work on the review – you've bagged **+10 points** this time! Here's the breakdown: **+18** as your base reward for being a reviewer, but we had to dock **-8** since there were no comments posted (the policy hits us with that penalty when reviews are totally silent). Your running score is now [+1727](https://zerocracy.github.io/judges-action/zerocracy-vitals.html) – pretty solid! Don't forget to peek at your [Zerocracy account](https://www.zerocracy.com/) when you get a chance.

0crat:
@vasilevnstas Thanks for the contribution! You've earned **+20 points** for this: **+24** as a basis; **-4** for contributing less than **30** hits-of-code (you had **29**). According to our [work policy](https://www.yegor256.com/2014/11/14/hits-of-code.html), aim for at least **30** hits-of-code to avoid this penalty and maximize your reward. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming! Your running score is [+1510](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #530, opened 2026-06-03
#526: Increase tombstone random _id range to avoid collisions

VasilevNStas:
In `lib/fbe/tombstone.rb:56`, `SecureRandom.random_number(99_999)` gives only ~100K possible values. For a distributed factbase, collisions are virtually guaranteed. Increased to `9_999_999_999_999` (10 trillion).

### DIFF
--- a/lib/fbe/tombstone.rb
+++ b/lib/fbe/tombstone.rb
@@ -56,7 +56,7 @@ class Fbe::Tombstone
         n.where = where
         n.repository = repo
       end
-    f.public_send(:"#{@fid}=", SecureRandom.random_number(99_999)) if f[@fid].nil?
+    f.public_send(:"#{@fid}=", SecureRandom.random_number(9_999_999_999_999)) if f[@fid].nil?
     nn = f['issues']&.map do |ii|
       ii.split('-').map do |i|
         Integer(i, 10)
--- a/test/fbe/test_tombstone.rb
+++ b/test/fbe/test_tombstone.rb
@@ -89,6 +89,19 @@ class TestTombstone < Fbe::Test
     assert_equal(%w[4-6 8 10-15 20], fb.query('(always)').each.first['issues'])
   end
 
+  def test_no_id_collisions
+    fb = Factbase.new
+    ts = Fbe::Tombstone.new(fb:)
+    n = 100
+    n.times { |i| ts.bury!('github', i, 1) }
+    ids = Set.new
+    fb.query("(and (eq what 'tombstone'))").each do |f|
+      refute_nil(f[:_id])
+      ids << f[:_id]
+    end
+    assert_equal(n, ids.size)
+  end
+
   def test_store_single_issues_without_turning_them_into_pairs
     where = 'github'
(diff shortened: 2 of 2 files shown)


VasilevNStas:
@yegor256  please review

yegor256:
@VasilevNStas you can easily create a test for this bug

VasilevNStas:
@yegor256 test added, please re-review

GHX5T-SOL (changes requested):
Thanks for adding the range increase. The production change looks right, and I verified `bundle exec rake test` plus `bundle exec rubocop` locally. However, I think the new regression test still needs to be deterministic before merge: `test_no_id_collisions` creates only 100 tombstones, so with the old 99,999-value range the collision probability is only about 4.83%. In other words, the test would pass on the buggy code roughly 95% of the time and would not reliably prevent this regression. Could you stub or control `SecureRandom.random_number`, or otherwise force an `_id` collision under the old cap, so the test fails before the fix and passes after it?

VasilevNStas:
Test rewritten to be deterministic: stubs SecureRandom.random_number and verifies it's called with `9_999_999_999_999`, instead of relying on 100 inserts with probabilistic collision detection.

GHX5T-SOL (approved):
Approved current head `dcc22745f19780f55012828f87be67409fa0c3dd`. The prior concern is addressed: `test_uses_large_random_range` stubs `SecureRandom.random_number` and asserts the call uses `9_999_999_999_999`, so the test deterministically fails against the old `99_999` cap and passes with this fix.

Validation run locally: `bundle check`; `git diff --check origin/master...HEAD`; Ruby syntax checks for `lib/fbe/tombstone.rb` and `test/fbe/test_tombstone.rb`; focused `bundle exec ruby -Itest test/fbe/test_tombstone.rb` (`8` tests, `32` assertions); touched-file `bundle exec rubocop lib/fbe/tombstone.rb test/fbe/test_tombstone.rb` (`2` files, no offenses); full `bundle exec rake test` (`284` tests, `804` assertions, `9` skips); and full `bundle exec rubocop` (`65` files, no offenses). Hosted checks are also green on the reviewed head.

0crat:
@ghx5t-sol Thanks for the review! You've earned **+6 points** for this contribution: **+18** as the base reward, **-8** for no comments posted during review, and **-4** for having fewer than 24 hits-of-code (22 total). Your running score is [+1158](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) as well.

0crat:
@vasilevnstas Great work on your contribution! 🎉 You've earned **+20 points** (+24 base, -4 for 22 hits-of-code). [Keep the momentum going](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) - you're doing amazing! Your total score is now [+1490](https://zerocracy.github.io/judges-action/zerocracy-vitals.html). Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #535, opened 2026-06-03
#527: Thread-safe global caching with GLOBAL_MUTEX

VasilevNStas:
Wrapped `global[:fb] ||=` (fb.rb), `global[:octo] ||=` (octo.rb), and `global[:github_graph] ||=` (github_graph.rb) in `Fbe::GLOBAL_MUTEX.synchronize` to prevent race conditions when multiple judges initialize these clients concurrently.

Bypasses the need for per-file lazy mutex init (avoids double-checked locking race on the mutex itself).

- Added `Fbe::GLOBAL_MUTEX = Mutex.new` to `lib/fbe.rb`
- All tests pass, RuboCop clean

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -108,6 +108,11 @@ Elegant/GoodMethodName:
     - workflow_run
     - workflow_run_job
     - workflow_run_usage
+Elegant/GoodVariableName:
+  AllowedNames:
+    - '@fb_mutex'
+    - '@octo_mutex'
+    - '@graph_mutex'
 Naming/MethodParameterName:
   AllowedNames:
     - as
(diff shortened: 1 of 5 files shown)


VasilevNStas:
@yegor256  please review

yegor256:
@VasilevNStas global state is pure evil in OOP: https://www.yegor256.com/2018/07/03/global-variables.html

VasilevNStas:
@yegor256 replaced GLOBAL_MUTEX with per-file Monitors hidden as module-level instance variables. No new public constants. Please re-review

yegor256:
@VasilevNStas this is essentially the same - still a global state

VasilevNStas:
Closing per @yegor256 feedback: module-level Monitors are still global state. Moved discussion to #527 to find an approach that avoids global state entirely.

PR #536, opened 2026-06-03
#529 fix stale cached state in Iterate#over on restart

VasilevNStas:
Made `values.delete(repo)` unconditional (previously guarded by `if @sorting`) so cached enumerator state is always cleared on restart for both sorting and non-sorting paths.

All tests pass, RuboCop clean.


VasilevNStas:
Closing this PR. The N+1 pattern in the non-sorting path is inherent to the incremental iteration design — see detailed explanation on #529.

PR #495, opened 2026-06-02
#474: Anchor Fbe.mask_to_regex regex to prevent prefix collisions

VasilevNStas:
## Description

`Fbe.mask_to_regex` builds a regex without `\A`/`\z` anchors, causing `re.match?` to match substrings. For exclusion masks like `-zerocracy/judges-action`, this means repositories with prefix collisions (e.g. `zerocracy/judges-actions` or `zerocracy/judges-action-old`) are incorrectly excluded.

## Fix

Added `\A` and `\z` anchors to the regex compilation in `lib/fbe/unmask_repos.rb:27`:

```ruby
Regexp.compile("\\A#{org}/#{repo.gsub('*', '.*')}\\z", Regexp::IGNORECASE)
```

Wildcard masks (`zerocracy/*`) are unaffected because `.*` already absorbs the entire tail.

## Tests

Added four direct unit tests for `Fbe.mask_to_regex` in `test/fbe/test_unmask_repos.rb`:

- `test_mask_to_regex_exact_match` — confirms exact match works
- `test_mask_to_regex_prefix_collision` — confirms `zerocracy/judges-action` does NOT match `zerocracy/judges-actions` or `zerocracy/judges-action-old`
- `test_mask_to_regex_wildcard` — confirms `zerocracy/*` still matches any repo
- `test_mask_to_regex_case_insensitive` — confirms case-insensitive flag is preserved

## Verification

- `bundle exec rubocop` — 0 offenses
- `bundle exec rake test` — 287 tests, 0 failures

Closes #474

PR #540, opened 2026-06-04
fix: #538 escape regex metacharacters in Fbe.mask_to_regex

edmoffo:
Fbe.mask_to_regex used to interpolate the mask straight into the pattern and only translate `*`. Every other regex metacharacter (a dot, a plus, parentheses, etc.) survived unchanged and was reinterpreted as an operator at match time. A wildcard like `zold-io/blog.zold.*` therefore matched siblings such as `zold-io/blogXzoldYio`, and a literal exclusion like `-zold-io/blog.zold.io` dropped repositories whose names merely shared that skeleton. Issue #538 lays the failure out file-by-file.

The change routes both halves of the mask through `Regexp.escape` and then turns the escaped literal `\*` back into `.*` so wildcards keep working. Only `lib/fbe/unmask_repos.rb:27` is touched, plus three minitest cases in `test/fbe/test_unmask_repos.rb` covering a literal dot, a non-dot metacharacter (`+`), and the wildcard-still-expands path.

Local checks against ruby 3.3.6 on linux: `bundle exec ruby -Itest test/fbe/test_unmask_repos.rb` runs the seven non-skipped tests with 24 assertions and 0 failures; `rubocop lib/fbe/unmask_repos.rb test/fbe/test_unmask_repos.rb --format simple` reports no offenses. The wider `rake test` shows 10 pre-existing `Fbe::OffQuota` errors in `test/fbe/test_octo.rb` and the middleware suites that also reproduce on a clean `master` checkout in this sandbox, so they are not introduced by this commit; GitHub CI on `master` is currently green, which matches the pre-existing-vs-environment split.

Closes #538

### DIFF
--- a/lib/fbe/unmask_repos.rb
+++ b/lib/fbe/unmask_repos.rb
@@ -24,7 +24,7 @@ require_relative 'over'
 def Fbe.mask_to_regex(mask)
   org, repo = mask.split('/')
   raise(Fbe::Error, "Org '#{org}' can't have an asterisk") if org.include?('*')
-  Regexp.compile("#{org}/#{repo.gsub('*', '.*')}", Regexp::IGNORECASE)
+  Regexp.compile("#{Regexp.escape(org)}/#{Regexp.escape(repo).gsub('\\*', '.*')}", Regexp::IGNORECASE)
 end
 
 # Resolves repository masks to actual GitHub repository names.
--- a/test/fbe/test_unmask_repos.rb
+++ b/test/fbe/test_unmask_repos.rb
@@ -50,6 +50,26 @@ class TestUnmaskRepos < Fbe::Test
     assert_equal(2, list.size)
   end
 
+  def test_mask_to_regex_treats_dot_as_literal
+    re = Fbe.mask_to_regex('zold-io/blog.zold.io')
+    assert_match(re, 'zold-io/blog.zold.io')
+    refute_match(re, 'zold-io/blogXzoldYio')
+  end
+
+  def test_mask_to_regex_escapes_other_metacharacters
+    re = Fbe.mask_to_regex('foo/bar+baz')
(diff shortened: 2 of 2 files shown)


kreinba (approved):
Routes both org and repo through Regexp.escape and then re-expands only the escaped asterisks via gsub('\\\\*', '.*'), which is exactly the minimal fix #538 prescribes. The three new tests pin the three behavioral paths: dot stays literal, plus stays literal, wildcard still matches. The anchoring concern from #474 is separately tracked in #495 as the issue body itself notes, and is out of scope here. CI is green.

0crat:
@kreinba Thanks for the review! You've earned **+6 points** for this contribution. Here's the breakdown per our [work policy](policy-link): **+18** base points for authoring the review, **-8** points deducted due to no comments being posted during the review process, and **-4** points deducted for having only 22 hits-of-code (below the 24 minimum threshold). Your running score is [+1550](https://zerocracy.github.io/judges-action/zerocracy-vitals.html) – keep up the great work and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates! 💪

0crat:
@edmoffo Thanks for the contribution! You've earned **+20 points** following our [work policy](https://www.yegor256.com/2014/11/14/hits-of-code.html): **+24** as the base award, then **-4** deducted for contributing fewer than 30 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (you had 22). To maximize your bonus next time, aim for 30+ hits-of-code to avoid this penalty. [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming! Your running score is [+1871](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #541, opened 2026-06-04
#539: Delete_one guard against fact recreation when value is not in array

VasilevNStas:
Closes #539.

When `Fbe.delete_one` is called with a value that is not present in the existing array property, the subtraction `before[prop] - [value]` returns the array unchanged. Previously the code still fell through to the `delete!` + `insert` cycle, destroying the fact and recreating it with the same content but a brand-new `_id`.

This is a sibling of #515 but with its own trigger: there the recreate path runs because a real change was requested; here it runs even though the array did not change, silently rewriting the fact identity and dangling any references to the original `_id`.

**Fix:** Early return when the subtraction produces the same array as the original.

**Test:** New `test_does_not_recreate_fact_when_value_not_present` builds a fact through `Fbe.fb` (so the `Pre` hook assigns `_id`), calls `delete_one` with a value not in the array, and asserts the `_id` is unchanged.

All tests pass, RuboCop clean.

### DIFF
--- a/lib/fbe/delete_one.rb
+++ b/lib/fbe/delete_one.rb
@@ -28,8 +28,10 @@ def Fbe.delete_one(fact, prop, value, fb: Fbe.fb, id: '_id')
     before[k] = fact[k]
   end
   return unless before[prop]
-  before[prop] = before[prop] - [value]
-  before.delete(prop) if before[prop].empty?
+  nv = before[prop] - [value]
+  return if nv == before[prop]
+  before[prop] = nv
+  before.delete(prop) if nv.empty?
   fb.query("(eq #{id} #{i})").delete!
   fb.txn do |fbt|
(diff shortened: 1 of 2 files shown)


VasilevNStas:
@yegor256  please review

0crat:
@vasilevnstas Thanks for the contribution! You've earned **+20 points** for this: +24 as a basis; -4 for contributing less than 30 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (22 total). Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+792](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #559, opened 2026-06-11
#515: preserve system properties on overwrite/delete_one recreate paths

bibonix:
Closes #515

The hash and scalar recreate paths in `lib/fbe/overwrite.rb` and the recreate path in `lib/fbe/delete_one.rb` already preserved every user-defined property, but the `Factbase::Pre` hook in `lib/fbe/fb.rb:43-49` reassigned `_id`, `_time`, `_version`, and `_job` on every `fbt.insert`, and the subsequent restore loop's `next unless n[k].nil?` guard then skipped putting the originals back. This patch applies the same drill-through pattern that `Fbe.delete` already uses (introduced for #542 in #545): walk the `@fact`/`@origin` decorator chain to the inner `Factbase::Fact`, delete the system keys from its `@map` whenever they were captured into `before`, and let the existing loop restore them.

The visible effect is that a fact recreated via `Fbe.overwrite` (scalar or hash form, overwrites-existing branch) or `Fbe.delete_one` against an `Fbe.fb`-decorated factbase now keeps the same `_id` (and `_time`, `_version`, `_job`), so external references to the original `_id` no longer dangle. Three regression tests in `test/fbe/test_overwrite.rb` and `test/fbe/test_delete_one.rb` build the factbase through `Fbe.fb`, snapshot the system properties before the call, and assert they equal the post-call values; they followed the model of `test_preserves_system_props_on_decorated_fb` already in `test/fbe/test_delete.rb`.

Checks run locally on ruby 3.3.6:
- `bundle exec ruby -Ilib -Itest test/fbe/test_overwrite.rb` -> 31 tests, 74 assertions, 0 failures, 0 errors, 0 skips
- `bundle exec ruby -Ilib -Itest test/fbe/test_delete_one.rb` -> 5 tests, 19 assertions, 0 failures, 0 errors, 0 skips
- `ruby -e 'require "rubocop"; RuboCop::CLI.new.run(["lib/fbe/delete_one.rb", "lib/fbe/overwrite.rb", "test/fbe/test_delete_one.rb", "test/fbe/test_overwrite.rb"])'` -> 4 files inspected, no offenses
- `git diff --check origin/master...HEAD` -> clean

### DIFF
--- a/lib/fbe/delete_one.rb
+++ b/lib/fbe/delete_one.rb
@@ -35,6 +35,13 @@ def Fbe.delete_one(fact, prop, value, fb: Fbe.fb, id: '_id')
   fb.query("(eq #{id} #{i})").delete!
   fb.txn do |fbt|
     c = fbt.insert
+    f = c
+    while f.instance_variable_defined?(:@fact) || f.instance_variable_defined?(:@origin)
+      iv = f.instance_variable_defined?(:@fact) ? :@fact : :@origin
+      f = f.instance_variable_get(iv)
+    end
+    map = f.instance_variable_get(:@map)
+    %w[_id _time _version _job].each { |k| map.delete(k) if before.key?(k) }
     before.each do |k, vv|
       next unless c[k].nil?
       vv.each do |v|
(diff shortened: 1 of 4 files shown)


0crat:
@bibonix Great work on this contribution! 🎉 You've earned **+24 points** following our policy: **+24** base points for the review, **+7.125** points for your 95 hits-of-code (capped at +16), and **-7** points to stay within the maximum 24-point limit. Your solid code size shows quality work without going overboard - [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming! Your running score is [+1091](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #570, opened 2026-06-26
#558 widen sqlite_store rescue to cover JSON::ParserError

edmoffo:
`Fbe::Middleware::SqliteStore#read` previously rescued only `Zlib::Error`, so a deflate stream that decoded into a non-JSON payload — partial writes, SQLite corruption recovery, leftover rows from older serializations — escaped as `JSON::ParserError` from `read` and crashed `Faraday::HttpCache` on every subsequent lookup for that key. The poisoned row stayed in the database because the existing `delete(key)` branch only ran on a Zlib failure, so until something else evicted the key the middleware was wedged on it.

The rescue is widened to also catch `JSON::ParserError` and `TypeError`, and the log line is renamed from `Failed to decompress` to `Failed to decode` to cover both branches. `delete(key)` still runs, so the next request rebuilds the entry from the network.

A regression test in `test/fbe/middleware/test_sqlite_store.rb` writes a valid deflate stream of a non-JSON payload directly through the store's SQLite handle and asserts that `read(key)` returns `nil` and that `all` reports an empty cache. Ran `bundle exec ruby -Ilib -Itest test/fbe/middleware/test_sqlite_store.rb` locally — 24 tests pass — and `bundle exec rubocop` is clean on both touched files.

Closes #558

---

### DIFF
--- a/lib/fbe/middleware/sqlite_store.rb
+++ b/lib/fbe/middleware/sqlite_store.rb
@@ -86,8 +86,8 @@ class Fbe::Middleware::SqliteStore
     return unless value
     begin
       JSON.parse(Zlib::Inflate.inflate(value))
-    rescue Zlib::Error => e
-      @loog.info("Failed to decompress cached value for key: #{key}, error: #{e.message}, the key will be deleted")
+    rescue Zlib::Error, JSON::ParserError, TypeError => e
+      @loog.info("Failed to decode cached value for key: #{key}, error: #{e.message}, the key will be deleted")
       delete(key)
     end
   end
--- a/test/fbe/middleware/test_sqlite_store.rb
+++ b/test/fbe/middleware/test_sqlite_store.rb
@@ -238,6 +238,21 @@ class SqliteStoreTest < Fbe::Test
     end
   end
 
+  def test_non_json_deflated_payload_does_not_raise
+    with_tmpfile('c.db') do |f|
+      Fbe::Middleware::SqliteStore.new(f, '0.0.1', loog: fake_loog).then do |store|
+        store.__send__(:perform) do |t|
+          t.execute(
+            'INSERT INTO cache(key, value, touched_at, created_at) VALUES(?1, ?2, ?3, ?3);',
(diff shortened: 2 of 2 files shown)


lib/fbe/middleware/sqlite_store.rb
```diff
@@ -86,8 +86,8 @@ def read(key)
     return unless value
     begin
       JSON.parse(Zlib::Inflate.inflate(value))
-    rescue Zlib::Error => e
-      @loog.info("Failed to decompress cached value for key: #{key}, error: #{e.message}, the key will be deleted")
+    rescue Zlib::Error, JSON::ParserError, TypeError => e
```
morphqdd, line 89: Good — this covers the full decode pipeline (inflate + parse), not just the inflate step, so a corrupted-but-inflatable entry no longer takes down the caller. `TypeError` also guards the non-string-inflate edge.

morphqdd (approved):
Correct fix for #558 — a cached value that inflates cleanly but isn't valid JSON previously raised an uncaught `JSON::ParserError`; widening the rescue to `JSON::ParserError`/`TypeError` (and generalizing the log wording from 'decompress' to 'decode') makes `read` self-heal by dropping the poisoned key. The test inserts a deflated non-JSON payload and asserts a clean nil + emptied cache. LGTM.

0crat:
@morphqdd Thanks for the review! You've earned +9 points for this: +18 as a basis; -5 for very few (1) comments; -4 for too few (19) hits-of-code.  Your running score is [+765](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@edmoffo Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (19) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+438](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #572, opened 2026-06-28
#565 Conclude#follow raises Fbe::Error on second call

VasilevNStas:
Fixes #565

`Conclude#follow` now raises `Fbe::Error` when called a second time, consistent with `by` and `as` in `Fbe::Iterate`.

Previously it silently overwrote `@follows`.

### DIFF
--- a/lib/fbe/conclude.rb
+++ b/lib/fbe/conclude.rb
@@ -123,6 +123,7 @@ class Fbe::Conclude
   # @param [Array<String>] props List of property names
   # @return [nil] Nothing
   def follow(props)
+    raise(Fbe::Error, 'Follow is already set') unless @follows.empty?
     @follows = props.strip.split.compact
   end
 
--- a/test/fbe/test_conclude.rb
+++ b/test/fbe/test_conclude.rb
@@ -205,6 +205,27 @@ class TestConclude < Fbe::Test
     assert_equal(%w[a b c], n['tags'])
   end
 
+  def test_follow_raises_on_second_call
+    $fb = Factbase.new
+    $global = {}
+    $epoch = Time.now
+    $loog = Loog::NULL
+    $options = Judges::Options.new
+    $fb.insert.foo = 1
+    assert_raises(Fbe::Error) do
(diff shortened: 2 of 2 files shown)


VasilevNStas:
@yegor256 please review

lib/fbe/conclude.rb
```diff
@@ -123,6 +123,7 @@ def on(query)
   # @param [Array<String>] props List of property names
   # @return [nil] Nothing
   def follow(props)
+    raise(Fbe::Error, 'Follow is already set') unless @follows.empty?
```
morphqdd, line 126: This behavior change ships without a test. Since #565 is specifically about the second call, a test that calls `follow` twice and asserts the second raises `Fbe::Error` (and that a single call still works) would lock in the new contract.

morphqdd (approved):
Implements #565 — `follow` now raises instead of silently overwriting on a second call, matching `by`. I checked `@follows` is initialized to `[]` in the constructor, so the first call's `@follows.empty?` is safe. One gap inline.

VasilevNStas:
Good catch. Added `test_follow_raises_on_second_call` — calls `follow` twice inside a `conclude` block and asserts the second raises `Fbe::Error`. Single-call path continues to work (verified by existing `test_follow_multivalued`).

0crat:
@morphqdd Thanks for the review! You've earned +9 points for this: +18 as a basis; -5 for very few (1) comments; -4 for too few (1) hits-of-code.  Your running score is [+698](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@vasilevnstas Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (22) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2000](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #573, opened 2026-06-28
#569 SqliteStore#close replaces per-instance at_exit handler

VasilevNStas:
Fixes #569

Removed the per-instance at_exit handler from perform that leaked database file descriptors and accumulated handlers in the test suite.

Added a public close method so callers can close the database connection deterministically. close also sets @db = nil, allowing re-open on next use.

Updated test_defer_db_close_callback to test_close_reopens_on_next_use to directly test explicit close instead of relying on subprocess at_exit behavior.

### DIFF
--- a/lib/fbe/middleware/sqlite_store.rb
+++ b/lib/fbe/middleware/sqlite_store.rb
@@ -161,6 +161,13 @@ class Fbe::Middleware::SqliteStore
     perform { _1.execute('SELECT key, value FROM cache') }
   end
 
+  # Close the database connection explicitly.
+  # @return [nil]
+  def close
+    @db&.close
+    @db = nil
+  end
+
   private
 
   def perform(&) # rubocop:disable Metrics/AbcSize
@@ -250,7 +257,6 @@ class Fbe::Middleware::SqliteStore
             "new file size: #{Filesize.from(File.size(@path).to_s).pretty} bytes"
           )
         end
-        at_exit { @db&.close }
       end
     @db.transaction(&)
   end
--- a/test/fbe/middleware/test_sqlite_store.rb
+++ b/test/fbe/middleware/test_sqlite_store.rb
@@ -3,9 +3,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
 # SPDX-License-Identifier: MIT
 
-require 'qbash'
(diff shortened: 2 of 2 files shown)


VasilevNStas:
@yegor256 plz review

lib/fbe/middleware/sqlite_store.rb
```diff
@@ -161,6 +161,13 @@ def all
     perform { _1.execute('SELECT key, value FROM cache') }
   end
 
+  # Close the database connection explicitly.
+  # @return [nil]
+  def close
```
morphqdd, line 166: Worth confirming intent: dropping `at_exit` means the connection is no longer closed automatically at process exit — callers must call `close` themselves. The OS reclaims the fd and SQLite in WAL mode is safe, so this is fine in practice; just make sure no caller relied on the implicit close for flushing.

morphqdd (approved):
Good fix for #569 — the `at_exit { @db&.close }` inside `perform` registered a handler that never got cleaned up; replacing it with an explicit `close` (plus lazy reopen via `@db ||=`) is the right shape, and the new test verifies close-then-reopen keeps the data. One note inline.

VasilevNStas:
Good point. Dropping `at_exit` is intentional:

1. SQLite in WAL mode ensures unflushed writes survive a crash — process exit without explicit close is safe.
2. The OS reclaims the file descriptor on exit regardless.
3. In production (`octo.rb`), the `SqliteStore` instance lives for the full process lifetime — no caller relied on the implicit close for flushing.
4. In tests, callers that need deterministic cleanup now call `close` explicitly.

0crat:
@morphqdd Thanks for the review! You've earned +13 points for this: +18 as a basis; -5 for very few (1) comments.  Your running score is [+711](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@vasilevnstas Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2024](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #594, opened 2026-07-07
Persist exhausted-cursor reset so Fbe::Iterate wraps around

yegor256:
`Fbe::Iterate#over` is meant to make rounds: once a repository's candidate list is exhausted, the cursor resets to `since` and the next run re-scans from the beginning. That reset never survived a run. The `ensure` block declined to write the marker whenever it equalled `since`, so the following run reloaded the stale high-water value and the cursor only ever climbed.

The effect was that any item becoming eligible after the cursor had passed it was orphaned forever. A pull request seen while still open advances the cursor past its number; if it merges later it sits below the cursor and is never re-examined, so `pull-was-merged` never fires and the work goes unrewarded.

The fix drops the `before[repo] == @since` guard so a genuine restart is persisted, while the `before[repo] == starts[repo]` guard still suppresses spurious writes when a repository made no progress. A new test locks in the wrap-around, and `test_sort_by_configuration` was updated to reflect the corrected marker values now that an exhausted repo rewinds to `since`.

Closes #593

### DIFF
--- a/lib/fbe/iterate.rb
+++ b/lib/fbe/iterate.rb
@@ -355,7 +355,7 @@ class Fbe::Iterate
        defined?(before) && !before.nil? &&
        defined?(starts) && !starts.nil?
       repos.each do |repo|
-        next if before[repo] == starts[repo] || before[repo] == @since
+        next if before[repo] == starts[repo]
         f =
           Fbe.if_absent(fb: @fb, always: true) do |n|
             n.what = 'iterate'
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +12 points for this: +24 as a basis; -8 for the lack of code review; -4 for too few (25) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+469](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #580, opened 2026-07-01
Fbe::Conclude#fill crashes with NoMethodError when follow property is missing on fact


sdghsdkjlas27-dotcom:
conclude.rb line 249 calls prev[follow.to_s].each. In Factbase, if a property doesn't exist on a fact, fact[property] returns nil. Calling .each on nil raises NoMethodError. If a query matches facts where some lack the follow properties, the entire conclude operation crashes mid-transaction.

REPRODUCTION: Use follow('nonexistent_property') with a query that matches facts lacking that property.

IMPACT: RuntimeError/NoMethodError during conclude; partial processing (some facts processed, some not); factbase may be left in inconsistent state.

VasilevNStas:
@yegor256 

This was fixed as a side-effect of #596 — the `next if values.nil?` guard in `Conclude#fill` (`lib/fbe/conclude.rb:280`) now skips missing followed properties instead of crashing. 

The issue can be closed.

0crat:
@sdghsdkjlas27-dotcom Thanks for reporting a new bug! You've earned +18 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+288](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

morphqdd:
This does not reproduce any more. `Conclude#fill` skips a property the fact does not carry:

```ruby
values = prev[key]
next if values.nil?
```

The guard came with #596. I ran the reproduction from the report, `follow` on a property no
fact has, and the conclude finishes and inserts its fact:

```ruby
Fbe.conclude(fb:, judge: 'j', ...) do
  on '(eq what "x")'
  follow 'nonexistent_property'
  draw { |n, _prev| n.what = 'y'; 'done' }
end
# no crash; facts now: 2
```

VasilevNStas:
@morphqdd thanks for confirming! Since the fix from #596 works and the issue no longer reproduces, could you (or @yegor256) please close it?

PR #596, opened 2026-07-07
fix: #595 honor (as ...) rewrite when following properties in Conclude#fill

yegor256:
Fixes #595.

## Problem

`Fbe::Conclude#fill` copied each followed property with `prev[follow.to_s]`. Factbase applies the `(as new old)` term rewrite only on **method access** (`prev.who`), not on `[]`, which returns the raw stored value. So following an `as`-renamed property copied the wrong value.

Minimal repro: for `(and (eq what 'x') (as who assignee))` where the fact has `who=777` and `assignee=4536`, `prev.who` returns `4536` but `prev['who']` returns `[777]`. `fill` was writing `777` into the concluded fact, silently corrupting every judge that follows an `as`-rewritten property (e.g. `bug-was-resolved`, `code-was-contributed`, `code-was-reviewed`).

The prior switch to `[]` (#520) was made to stop truncating multi-valued follows (#509), but `[]` bypasses `(as …)`. Neither accessor is correct alone.

## Fix

In `fill`, for each followed property:

1. Read the raw values via `prev[key]`; skip the property if it's absent (`nil`) instead of crashing on `nil.each` (also addresses the same-line crash in #580).
2. Read the rewritten value via method access (`prev.public_send(follow)`), which honors `(as …)`.
3. When the rewritten value differs from the first raw value — i.e. a rewrite is in effect — use the rewritten value; otherwise copy every raw value, preserving the multi-valued behavior from #520.

## Tests

- `test_follow_honors_as_rewrite` — `(as who assignee)` now copies `4536`, not `777`.
- `test_follow_skips_missing_property` — following an absent property no longer crashes.
- Existing `test_follow_multivalued` still passes (all values copied for genuine multi-valued follows).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Qq9uDEo7ckQ959pCdiDpei

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qq9uDEo7ckQ959pCdiDpei)_

### DIFF
--- a/lib/fbe/conclude.rb
+++ b/lib/fbe/conclude.rb
@@ -231,6 +231,13 @@ class Fbe::Conclude
   # calls the provided block for custom processing, and sets metadata
   # on the new fact.
   #
+  # A property that is absent on the previous fact is skipped (its "[]"
+  # accessor returns nil). Factbase's "(as new old)" rewrite is only visible
+  # through method access (prev.new), while "[]" returns the raw stored value
+  # (see #595); when such a rewrite is in effect the method accessor returns
+  # the rewritten value, so we honor it. All values are still copied for
+  # genuinely multi-valued properties (see #520).
+  #
   # @param [Factbase::Fact] fact The fact to populate
   # @param [Factbase::Fact] prev The previous fact to copy from
   # @yield [Factbase::Fact, Factbase::Fact] New fact and the previous fact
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +16 points for this: +24 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+485](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #599, opened 2026-07-08
fix: #597 reserve a per-run slot to stop before a step overruns the timeout

yegor256:
Closes #597.

## Problem

`Fbe.consider`/`Fbe.draw` (via the shared `roll` loop in `lib/fbe/conclude.rb`) only stop iterating when `Fbe.over?` reports the budget spent — a fixed 10% margin before the per-judge timeout (`timeout * 0.9`). For judges whose every iteration makes several network calls (e.g. `zerocracy/j`'s `pay-reward`), a step that starts near the end of the budget overruns, and the `judges` gem's hard `Timeout.timeout(options.timeout)` kills the judge. The loop-level check can't help once a slow step is already in flight.

## Change

- Add an integer `slot:` parameter (seconds needed per run, default `1`) to `Fbe.conclude` and the `Fbe::Conclude` constructor, plumbed into `roll`.
- In the loop, right after the existing `break if Fbe.over?(...)`, also break when less than one slot remains in the timeout budget (`options.timeout - (Time.now - kickoff) < slot`) and likewise against the lifetime/epoch budget. The `timeout_unaware`/`lifetime_unaware` flags are respected.
- Default `slot: 1` is backward-safe, since `over?` already stops earlier at `timeout * 0.9`; only judges that raise `slot` change behavior.

## Tests

Added `test_slot_stops_before_timeout_overrun`, `test_slot_stops_before_lifetime_overrun`, and `test_default_slot_does_not_stop_early`. Full `test_conclude.rb` suite and RuboCop pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015YnD78F7h6QWHvXYnXWS8U

---
_Generated by [Claude Code](https://claude.ai/code/session_015YnD78F7h6QWHvXYnXWS8U)_

### DIFF
--- a/lib/fbe/conclude.rb
+++ b/lib/fbe/conclude.rb
@@ -19,17 +19,20 @@ require_relative 'over'
 # @param [Loog] loog The logging facility
 # @param [Time] epoch When the entire update started
 # @param [Time] kickoff When the particular judge started
+# @param [Integer] slot Seconds a single iteration may need; the loop stops
+#   before starting a new one when less than this many seconds remain in the
+#   timeout (or lifetime) budget
 # @yield [Factbase::Fact] The fact
 def Fbe.conclude(
   fb: Fbe.fb, judge: $judge, loog: $loog, options: $options, global: $global,
-  epoch: $epoch || Time.now, kickoff: $kickoff || Time.now, &
+  epoch: $epoch || Time.now, kickoff: $kickoff || Time.now, slot: 1, &
 )
   raise(Fbe::Error, 'The fb is nil') if fb.nil?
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +21 points for this: +24 as a basis; +5 for the 73 hits-of-code that you wrote; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+495](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #618, opened 2026-07-13
#616: deduplicate repos matched by overlapping masks in unmask_repos

### LINKED ISSUE #616: `Fbe.unmask_repos` yields the same repository twice when an exact-name mask overlaps a wildcard mask
`Fbe.unmask_repos` (`lib/fbe/unmask_repos.rb:71-104`) builds the `repos` array by iterating every non-exclusion mask independently and appending matches, with no deduplication anywhere in the pipeline (not after the inclusion loop, not after the exclusion loop, not before the final `repos.each`/return).

**What happens:** when `options.repositories` contains an exact repo name and a wildcard mask that also matches that same repo — e.g. `"yegor256/factbase,Yegor256/*"` — the repo is appended once by the exact-name branch (`repos << mask; next`) and again by the wildcard branch (`re.match?` loop over `organization_repositories`), and both copies survive to the end.

Reproduced directly:

```ruby
opts = Judges::Options.new({'testing' => true, 'repositories' => 'yegor256/factbase,Yegor256/*'})
list = Fbe.unmask_repos(options: opts, global: {}, loog: Loog::NULL)
# => ["yegor256/judges", "yegor256/factbase", "yegor256/factbase"]
list.size      # => 3
list.uniq.size # => 2
```

`yegor256/factbase` appears twice. Since the method's block form does `repos.each { |repo| ...; yield(repo) }`, a caller using overlapping masks gets the same repository yielded (and processed — extra API calls, extra judge-fact writes, etc.) twice per run for no legitimate reason.

**What should happen:** `repos` should contain each matching repository exactly once regardless of how many masks (exact or wildcard) happen to match it — e.g. `repos.uniq!` before the emptiness check / return.

morphqdd:
Closes #616.

`Fbe.unmask_repos` appended matches from every inclusion mask independently, so a repository matched by both an exact-name mask and a wildcard mask (e.g. `yegor256/factbase,Yegor256/*`) survived to the final list twice and was yielded — and processed — twice per run. A `repos.uniq!` now runs right after the exclusion pass, before the archived-repos filter, so duplicates also don't cost an extra `octo.repository` call each.

The new test reproduces the overlap from the issue and asserts the list is duplicate-free while still containing the doubly-matched repo; it fails on master.

Tests: 13 tests, 0 failures; rubocop: no offenses.

### DIFF
--- a/lib/fbe/unmask_repos.rb
+++ b/lib/fbe/unmask_repos.rb
@@ -92,6 +92,7 @@ def Fbe.unmask_repos( # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticCompl
     re = Fbe.mask_to_regex(mask[1..])
     repos.reject! { |r| re.match?(r) }
   end
+  repos.uniq!
   repos.reject! { |repo| octo.repository(repo)[:archived] }
   raise(Fbe::Error, "No repos found matching: #{options.repositories.inspect}") if repos.empty?
   repos.shuffle!
--- a/test/fbe/test_unmask_repos.rb
+++ b/test/fbe/test_unmask_repos.rb
@@ -50,6 +50,13 @@ class TestUnmaskRepos < Fbe::Test
     assert_equal(2, list.size)
   end
 
+  def test_deduplicates_repos_matched_by_overlapping_masks
+    opts = Judges::Options.new({ 'testing' => true, 'repositories' => 'yegor256/factbase,Yegor256/*' })
+    list = Fbe.unmask_repos(options: opts, global: {}, loog: Loog::NULL)
+    assert_equal(list.uniq.size, list.size, "duplicates found in #{list.inspect}")
+    assert_includes(list, 'yegor256/factbase')
+  end
+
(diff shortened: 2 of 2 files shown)


morphqdd:
@yegor256, please take a look when you have a moment.

morphqdd:
@yegor256, please take a look when you have a moment.

morphqdd:
@yegor256, review please

morphqdd:
@yegor256, please take a look when you have a moment.

yegor256:
Thanks for the contribution!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+668](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

